PR #154, opened 2021-02-17
glass: Use inventory

Devp00l:
Use the inventory API call to list all volumes on dashboard.


jecluis (changes requested):
I'm pretty sure this needs a

`/* eslint-disable @typescript-eslint/naming-convention */`

to suppress the snake-case warnings.

votdev:
> I'm pretty sure this needs a
> 
> `/* eslint-disable @typescript-eslint/naming-convention */`
> 
> to suppress the snake-case warnings.

Will fix that in a follow-up PR.

votdev:
> I'm pretty sure this needs a
> 
> `/* eslint-disable @typescript-eslint/naming-convention */`
> 
> to suppress the snake-case warnings.

The file has already disabled that rule.

PR #156, opened 2021-02-17
glass: Services dashboard widget

Devp00l:
Simple widget to only currently show the name and the type of the share
as some useful information is missing in the current API. It will route
to the deployment page if no service is found.

### DIFF
--- a/src/glass/src/app/shared/services/api/services.service.spec.ts
+++ b/src/glass/src/app/shared/services/api/services.service.spec.ts
@@ -1,12 +1,15 @@
+import { HttpClientTestingModule } from '@angular/common/http/testing';
 import { TestBed } from '@angular/core/testing';
 
-import { ServicesService } from '~/app/shared/services/api/services.service';
+import { ServicesService } from './services.service';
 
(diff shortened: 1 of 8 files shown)


src/glass/src/app/shared/services/api/services.service.spec.ts
```diff
@@ -1,12 +1,15 @@
+import { HttpClientTestingModule } from '@angular/common/http/testing';
 import { TestBed } from '@angular/core/testing';
 
-import { ServicesService } from '~/app/shared/services/api/services.service';
+import { ServicesService } from './services.service';
```
jecluis, line 4: shouldn't this be an absolute path?

jecluis (commented):
lgtm, but haven't tested it to check how it looks or behaves.

src/glass/src/app/shared/services/api/services.service.spec.ts
votdev: Will fix that in a follow-up PR.

votdev (approved):

PR #163, opened 2021-02-17
gravel: fix regression introduced by 9289ce3

jecluis:


mgfritch (approved):
*sigh* we really could use a few gstate tests ;)

PR #172, opened 2021-02-18
doc: adding extremely important badges to readme

asettle:
Signed-off-by: Alexandra Settle <asettle@suse.com>


jecluis (approved):
pretty!

PR #179, opened 2021-02-18
gravel: expose bootstrap progress

jecluis:
This is extremely fragile, as we're essentially parsing cephadm's
free-form output and associating each step with a rule of thumbs
percentage. Should cephadm change their messaging, this can easily
break.

### DIFF
--- a/src/gravel/cephadm/cephadm.py
+++ b/src/gravel/cephadm/cephadm.py
@@ -16,15 +16,25 @@ from logging import Logger
 import os
 import json
 from io import StringIO
-from typing import List, Tuple
+from typing import (
+    Callable,
+    List,
+    Optional,
+    Tuple
+)
 
 import pydantic
 from pydantic.tools import parse_obj_as
 from fastapi.logger import logger as fastapi_logger
 
-from .models import HostFactsModel, NodeCPUInfoModel, \
-    NodeCPULoadModel, NodeInfoModel, NodeMemoryInfoModel, \
+from .models import (
+    HostFactsModel,
+    NodeCPUInfoModel,
+    NodeCPULoadModel,
+    NodeInfoModel,
+    NodeMemoryInfoModel,
     VolumeDeviceModel
+)
 
 
 logger: Logger = fastapi_logger
@@ -47,7 +57,9 @@ class Cephadm:
 
         pass
 
-    async def call(self, cmd: str) -> Tuple[str, str, int]:
+    async def call(self, cmd: str,
+                   outcb: Optional[Callable[[str], None]] = None
+                   ) -> Tuple[str, str, int]:
 
         cmdlst: List[str] = f"{self.cephadm} {cmd}".split()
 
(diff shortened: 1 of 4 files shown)


votdev (approved):
LGTM and a great improvement.

votdev:
Can't we implement the feature to report the deployment stage in cephadm and contribute it upstream?

src/gravel/cephadm/cephadm.py
```diff
@@ -64,24 +66,50 @@ async def call(self, cmd: str) -> Tuple[str, str, int]:
...
+            ("creating mon", 20),
+            ("waiting for mon to start", 25),
+            ("wrote config", 40),
+            ("creating mgr", 45),
+            ("waiting for mgr to start", 50),
+            ("enabling cephadm module", 60),
+            ("setting orchestrator backend to cephadm", 70),
+            ("adding host", 80),
+            ("bootstrap complete", 100)
+        ]
```
mgfritch, line 103: super fragile, seems like this would justify a unit test with various types of output from cephadm ..

mgfritch (commented):
> Can't we implement the feature to report the deployment stage in cephadm and contribute it upstream?

I was wondering about the exact same thing ... not a huge fan of parsing arbitrary output.

jecluis:
> > Can't we implement the feature to report the deployment stage in cephadm and contribute it upstream?
> 
> I was wondering about the exact same thing ... not a huge fan of parsing arbitrary output.

@sebastian-philipp Thoughts? Would this be something that would be interesting upstream?

sebastian-philipp:
This needs serious refactoring in upstream. We have IMO to abandon the "single big blob of python executable" in cephadm first. 

1. https://github.com/ceph/ceph/pull/39141 (having reviews will help)
2. https://github.com/ceph/ceph/pull/37495
3. change the url in https://docs.ceph.com/en/latest/cephadm/install/#install-cephadm to download.ceph.com
4. make cephadm a proper module and zip/package/compile it into an executable. 

And then we can start with this. 

Doing this now would make future refactoring even more complicated and cumbersome.

jecluis:
@sebastian-philipp Ok, so... how can we help make that happen? Is there a path? Issues, tasks, something we can devote people and time to?

sebastian-philipp:
> @sebastian-philipp Ok, so... how can we help make that happen? Is there a path? Issues, tasks, something we can devote people and time to?

Reviewing https://github.com/ceph/ceph/pull/39141 would help me a lot!

sebastian-philipp:
next steps would be to upload cephadm to download.ceph.com instead of letting users wget from the source tree.

sebastian-philipp:
and please review https://github.com/ceph/ceph/pull/39619

jhesketh (approved):
lgtm except for missing tests

src/gravel/cephadm/cephadm.py
sebastian-philipp: https://github.com/ceph/ceph/pull/40103

jecluis: oof. Glad it didn't change any of those, eh? :)

jecluis:
@votdev rebased and repushed. Needs a green checkmark to merge.

votdev (approved):

PR #183, opened 2021-02-19
gravel: reduce Ticker boilerplate code

### LINKED ISSUE #143: gravel: add `tick` boilerplate as a decorator
All current (and likely future) implementations of `Ticker::tick()` do common work. For now that is
- check if its time to tick
- check if already ticking
- set ticking
- unset ticking after work is done

As this is already diverging with two implementations it makes sense to implement this in a decorator, that the various `tick` implementations can use.

jecluis:
Resolves #143

### DIFF
--- a/src/gravel/controllers/resources/inventory.py
+++ b/src/gravel/controllers/resources/inventory.py
@@ -20,7 +20,7 @@ class Inventory(Ticker):
     def __init__(self):
         super().__init__(
             "inventory",
-            gstate.config.options.inventory_probe_interval
+            gstate.config.options.inventory.probe_interval
         )
         self._latest = None
 
--- a/src/gravel/controllers/resources/inventory.py
+++ b/src/gravel/controllers/resources/inventory.py
@@ -18,28 +18,17 @@ class Inventory(Ticker):
     _latest: Optional[NodeInfoModel]
 
     def __init__(self):
-        self.is_ticking = False
+        super().__init__(
+            "inventory",
+            gstate.config.options.inventory_probe_interval
+        )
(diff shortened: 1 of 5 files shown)


src/gravel/controllers/resources/inventory.py
```diff
@@ -18,28 +18,17 @@ class Inventory(Ticker):
     _latest: Optional[NodeInfoModel]
 
     def __init__(self):
-        self.is_ticking = False
+        super().__init__(
+            "inventory",
+            gstate.config.options.inventory_probe_interval
```
mgfritch, line 23:
```suggestion
            gstate.config.options.inventory.probe_interval
```

nit, but wondering if we should split the `probe_interval` options by class?

jecluis:
@mgfritch repushed

src/gravel/controllers/tests/test_config.py
```diff
@@ -13,7 +13,7 @@ def setUp(self):
 
     def test_config_version(self):
         config = Config()
-        assert config.config.version == 2
+        assert config.config.version == 3
```
mgfritch, line 16: nit, but wondering at what point we should add various samples of config versions to the test ...

mgfritch (approved):
lgtm!

src/gravel/controllers/tests/test_config.py
jecluis: To be honest, this could have stayed the same all this time. This will only matter, I think, once we start trying to ensure backwards compatibility. I'd say once we cut the first release.

PR #186, opened 2021-02-19
gravel: Add check before calling realpath

votdev:
This will prevent the following error:
```
node01:/srv/aquarium # ./run_aquarium.sh --debug
realpath: missing operand
Try 'realpath --help' for more information.
```

### DIFF
--- a/run_aquarium.sh
+++ b/run_aquarium.sh
@@ -45,7 +45,9 @@ $has_config && [[ -z "${config_path}" ]] && \
   echo "error: must specify a PATH with '--config'" && \
   exit 1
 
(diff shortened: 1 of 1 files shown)


mgfritch (approved):
this works, but wondering if we even need to determine the realpath ..

jecluis (approved):

jecluis:
> this works, but wondering if we even need to determine the realpath ..

We do, otherwise relative paths at time of calling the script end up as relative paths at time of running uvicorn. I.e., `./run_aquarium.sh -c foo`, ends up in `src/foo/` instead of `./foo/`.

Issue #190, opened 2021-02-19
systemd: aquarium.service installed in wrong directory


l-mb:
aquarium installs its unit file under `/etc/systemd/system/`, which is reserved for local/admin overrides and customization.

Packages should install under `/usr/lib/systemd/system/`

PR #191, opened 2021-02-19
tools/build-image.sh: move systemd unit dir

mgfritch:
'/etc/systemd/system/' -> '/usr/lib/systemd/system/'

Resolves: #190


l-mb (approved):
LGTM!

PR #195, opened 2021-02-19
gravel: add unit test for config options

ml8mr:
Test inventory and storage probe intervals, plus the location of the
service state file (storage.json).

This test exposed a bug where the service state file location is not changing
based on parameters passed when initializing the Config object. Pytest reports
the problem as follows:

>       assert config.options.service_state_path == 'foo/storage.json'
E       AssertionError: assert '/etc/aquarium/storage.json' == 'foo/storage.json'
E         - foo/storage.json
E         + /etc/aquarium/storage.json

@mgfritch provided the fix for this issue in controllers/config.py.

Signed-off-by: Mike Latimer <mlatimer@suse.com>

### DIFF
--- a/src/gravel/controllers/config.py
+++ b/src/gravel/controllers/config.py
@@ -38,8 +38,8 @@ class StorageOptionsModel(BaseModel):
 
 
 class OptionsModel(BaseModel):
-    service_state_path: str = Field(Path(config_dir).joinpath("storage.json"),
-                                    title="Path to Service State file")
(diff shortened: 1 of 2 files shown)


jecluis (dismissed):

jecluis:
@mgfritch @ml8mr pull request is as WIP. Anything else to be pushed? It looks good to me for a merge.

mgfritch:
@jecluis added additional test where config.yaml and storage.yaml reside in different basedirs

jecluis (approved):

Issue #185, opened 2021-02-19
glass: add a polling service


s0nea:
We're doing API polling on some places in our installer. We might want to summarize them in one centralized service.

PR #197, opened 2021-02-22
glass: add poll service

s0nea:
Fixes: https://github.com/aquarist-labs/aquarium/issues/185

### DIFF
--- a/src/glass/src/app/pages/deployment-page/deployment-page.component.ts
+++ b/src/glass/src/app/pages/deployment-page/deployment-page.component.ts
@@ -2,14 +2,13 @@ import { Component, OnInit, ViewChild } from '@angular/core';
 import { MatDialog } from '@angular/material/dialog';
 import { MatStepper } from '@angular/material/stepper';
 import { BlockUI, NgBlockUI } from 'ng-block-ui';
-import { of } from 'rxjs';
-import { delay, mergeMap } from 'rxjs/operators';
 
 import { CephfsModalComponent } from '~/app/pages/deployment-page/cephfs-modal/cephfs-modal.component';
 import { DialogComponent } from '~/app/shared/components/dialog/dialog.component';
 import { Device, OrchService } from '~/app/shared/services/api/orch.service';
 import { ServiceDesc, ServicesService } from '~/app/shared/services/api/services.service';
 import { NotificationService } from '~/app/shared/services/notification.service';
+import { PollService } from '~/app/shared/services/poll.service';
 
 @Component({
   selector: 'glass-deployment-page',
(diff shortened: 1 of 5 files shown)


Devp00l (dismissed):
LGTM so far :D

src/glass/src/app/pages/deployment-page/deployment-page.component.ts
```diff
@@ -37,43 +36,47 @@ export class DeploymentPageComponent implements OnInit {
...
+    this.startBlockUI('Please wait, fetching device information ...');
+    this.orchService
+      .devices()
       .pipe(
-        delay(5000),
-        mergeMap(() => this.orchService.devices())
+        this.pollService.poll(
+          (hostDevices) => {
+            let hasDevices = false;
+            Object.values(hostDevices).forEach((v) => {
```
votdev, line 60: I think this should also work.
```suggestion
            let hasDevices = Object.values(hostDevices).some((v) => v.devices.length);
```

src/glass/src/app/pages/deployment-page/deployment-page.component.ts
```diff
@@ -103,32 +106,26 @@ export class DeploymentPageComponent implements OnInit {
...
 
   pollAssimilationStatus(): void {
-    of(true)
-      .pipe(
-        delay(5000),
-        mergeMap(() => this.orchService.assimilateStatus())
-      )
+    this.orchService
+      .assimilateStatus()
+      .pipe(this.pollService.poll((res) => !res, undefined, 'Failed to deploy disks.'))
```
votdev, line 121: We are using `devices` allover the page, so we should use it here, too.
```suggestion
      .pipe(this.pollService.poll((res) => !res, undefined, 'Failed to deploy devices.'))
```

votdev (changes requested):
Works, but need some adaptions.

src/glass/src/app/pages/deployment-page/deployment-page.component.ts
```diff
@@ -37,43 +36,39 @@ export class DeploymentPageComponent implements OnInit {
...
-    of(true)
+  getDevices(): void {
+    this.startBlockUI('Please wait, fetching device information ...');
+    this.orchService
+      .devices()
       .pipe(
-        delay(5000),
-        mergeMap(() => this.orchService.devices())
+        this.pollService.poll(
+          (hostDevices) => Object.values(hostDevices).some((v) => v.devices.length),
```
votdev, line 58:
```suggestion
          (hostDevices): boolean => !Object.values(hostDevices).some((v) => v.devices.length),
```

votdev (changes requested):
I would also suggest to move the service to `/shared/services` because it is not related to API only.

s0nea:
> I would also suggest to move the service to `/shared/services` because it is not related to API only.

Yeah... I put it into the `api` folder because the service can't be used outside of it... But if there was/will be a poll API endpoint, it probably makes sense to move it outside.

votdev (approved):

PR #206, opened 2021-02-22
gravel: pass the ceph.conf file to the Mgr/Mon class

mgfritch:
location to the `ceph.conf` file can be passed to the `Ceph` baseclass,
extend this to the Mgr/Mon classes

### DIFF
--- a/src/gravel/controllers/orch/ceph.py
+++ b/src/gravel/controllers/orch/ceph.py
@@ -11,6 +11,9 @@ from pathlib import Path
 from typing import Callable, Dict, Any, List
 
 
+CEPH_CONF_FILE = '/etc/ceph/ceph.conf'
+
+
 class CephError(Exception):
     pass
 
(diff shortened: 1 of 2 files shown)


src/gravel/controllers/orch/ceph.py
```diff
@@ -108,17 +111,17 @@ def call(self, cmd: Dict[str, Any]) -> Any:
 
 class Mgr(Ceph):
 
-    def __init__(self):
-        super().__init__()
+    def __init__(self, conf_file: str = CEPH_CONF_FILE):
```
sebastian-philipp, line 114: fyi, mons and mgrs don't use the `"/etc/ceph/ceph.conf"` at all. The file is just there for admin's convenience

src/gravel/controllers/orch/ceph.py
```diff
@@ -11,6 +11,9 @@
 from typing import Callable, Dict, Any, List
 
 
+CEPH_CONF_FILE = '/etc/ceph/ceph.conf'
```
sebastian-philipp, line 14: https://github.com/ceph/ceph/blob/a3591378a59c621ac597987facb3fb30707c218f/src/pybind/mgr/cephadm/serve.py#L346-L368

src/gravel/controllers/orch/ceph.py
mgfritch: eh but these are mon clients via rados .. unless I'm misunderstanding what you mean?

jecluis (approved):

PR #212, opened 2021-02-23
glass: Show real data in capacity widget

### LINKED ISSUE #210: glass: Show real data in capacity widget
**Is your feature request related to a problem? Please describe.**
Currently the capacity widget only shows mocked data and it also doesn't look so great.

**Describe the solution you'd like**
A solution to show the current capacity in a much cooler way would be to use https://swimlane.github.io/ngx-charts/#/ngx-charts/advanced-pie-chart as the user will spot instantly how much space he has and how much percentage is in use.

**Describe alternatives you've considered**
I also have thought about other charts and already tested how they would look, but the only intresting alternative would be the tree map (https://swimlane.github.io/ngx-charts/#/ngx-charts/tree-map).

Devp00l:
Fixes #210

capacity_real_data


votdev (approved):
That widget looks really cool now.

PR #225, opened 2021-02-28
gravel: initial support for node communication over websockets

jecluis:
Please note:

1. Existing deployments will be broken by these changes;
2. There's been a deployment stage format change, and we are now using integers instead of strings for the enum; @votdev mind looking at the glass bits?
3. Currently missing tests because I'm not fully satisfied with the direction, and don't want to be writing and rewriting tests until I'm sure the architecture makes sense.

### DIFF
--- a/src/glass/src/app/shared/services/api/status.service.ts
+++ b/src/glass/src/app/shared/services/api/status.service.ts
@@ -22,11 +22,20 @@ export interface ClusterStatus {
   health: HealthStatus;
 }
 
+// eslint-disable-next-line no-shadow
+export enum StatusStageEnum {
+  unknown = -1,
+  none = 0,
+  bootstrapping = 1,
+  bootstrapped = 2,
+  ready = 3,
+}
+
 export type Status = {
   /* eslint-disable @typescript-eslint/naming-convention */
   deployment_state: {
     last_modified: string;
-    stage: 'none' | 'bootstrapping' | 'bootstrapped' | 'ready' | 'unknown';
+    stage: StatusStageEnum;
   };
   cluster?: ClusterStatus;
 };
--- a/src/aquarium.py
+++ b/src/aquarium.py
@@ -31,6 +31,7 @@ from gravel.api import bootstrap
 from gravel.api import orch
 from gravel.api import status
 from gravel.api import services
+from gravel.api import nodes
 
 
 logger: logging.Logger = fastapi_logger
@@ -62,6 +63,7 @@ api.include_router(bootstrap.router)
 api.include_router(orch.router)
 api.include_router(status.router)
 api.include_router(services.router)
+api.include_router(nodes.router)
 
 
 #
--- a/src/glass/src/app/pages/bootstrap-page/bootstrap-page.component.ts
+++ b/src/glass/src/app/pages/bootstrap-page/bootstrap-page.component.ts
@@ -5,6 +5,7 @@ import { BlockUI, NgBlockUI } from 'ng-block-ui';
 import {
   BootstrapBasicReply,
   BootstrapService,
+  BootstrapStageEnum,
   BootstrapStatusReply
 } from '~/app/shared/services/api/bootstrap.service';
 import { NotificationService } from '~/app/shared/services/notification.service';
@@ -33,12 +34,12 @@ export class BootstrapPageComponent implements OnInit {
     // Immediately block the UI if bootstrapping is in progress.
     this.bootstrapService.status().subscribe({
       next: (statusReply: BootstrapStatusReply) => {
-        if (statusReply.stage === 'running') {
+        if (statusReply.stage === BootstrapStageEnum.running) {
           this.visible = false;
           this.blockUI.start('Please wait, bootstrapping in progress ...');
           this.pollBootstrapStatus();
         }
-        if (statusReply.stage === 'none') {
+        if (statusReply.stage === BootstrapStageEnum.none) {
           this.visible = true;
         }
       },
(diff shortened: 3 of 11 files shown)


src/glass/src/app/shared/services/api/status.service.ts
```diff
@@ -22,11 +22,19 @@ export interface ClusterStatus {
   health: HealthStatus;
 }
 
+export enum StageEnum {
```
votdev, line 25: I would suggest to use `StatusStageEnum` because we also have a `StageEnum` in the `/api/bootstrapt/status` response. Normally it is no problem having to types with the same name in TS because you can rename them on import, but IMO it is better to have unique names if possible.

```suggestion
export enum StatusStageEnum {
```

votdev (changes requested):
Please apply the following patch to get the UI working finally.
```
diff --git a/src/glass/src/app/pages/bootstrap-page/bootstrap-page.component.ts b/src/glass/src/app/pages/bootstrap-page/bootstrap-page.component.ts
--- a/src/glass/src/app/pages/bootstrap-page/bootstrap-page.component.ts	(revision 91e69960435cf1da7360f526f5829f9c3b3573ab)
+++ b/src/glass/src/app/pages/bootstrap-page/bootstrap-page.component.ts	(date 1614599263417)
@@ -5,8 +5,8 @@
 import {
   BootstrapBasicReply,
   BootstrapService,
-  BootstrapStatusReply
-} from '~/app/shared/services/api/bootstrap.service';
+  BootstrapStageEnum,
+  BootstrapStatusReply} from '~/app/shared/services/api/bootstrap.service';
 import { NotificationService } from '~/app/shared/services/notification.service';
 import { PollService } from '~/app/shared/services/poll.service';

@@ -33,12 +33,12 @@
     // Immediately block the UI if bootstrapping is in progress.
     this.bootstrapService.status().subscribe({
       next: (statusReply: BootstrapStatusReply) => {
-        if (statusReply.stage === 'running') {
+        if (statusReply.stage === BootstrapStageEnum.running) {
           this.visible = false;
           this.blockUI.start('Please wait, bootstrapping in progress ...');
           this.pollBootstrapStatus();
         }
-        if (statusReply.stage === 'none') {
+        if (statusReply.stage === BootstrapStageEnum.none) {
           this.visible = true;
         }
       },
@@ -82,7 +82,7 @@
       .status()
       .pipe(
         this.pollService.poll(
-          (statusReply) => statusReply.stage === 'running',
+          (statusReply) => statusReply.stage === BootstrapStageEnum.running,
           undefined,
           'Failed to bootstrap the system.'
         )
@@ -90,11 +90,11 @@
       .subscribe(
         (statusReply: BootstrapStatusReply) => {
           switch (statusReply.stage) {
-            case 'error':
+            case BootstrapStageEnum.error:
               handleError();
               break;
-            case 'none':
-            case 'done':
+            case BootstrapStageEnum.none:
+            case BootstrapStageEnum.done:
               this.blockUI.stop();
               this.router.navigate(['/installer/create/deployment']);
               break;
===================================================================
diff --git a/src/glass/src/app/shared/services/api/status.service.ts b/src/glass/src/app/shared/services/api/status.service.ts
--- a/src/glass/src/app/shared/services/api/status.service.ts	(revision 91e69960435cf1da7360f526f5829f9c3b3573ab)
+++ b/src/glass/src/app/shared/services/api/status.service.ts	(date 1614599263465)
@@ -22,6 +22,7 @@
   health: HealthStatus;
 }

+// eslint-disable-next-line no-shadow
 export enum StageEnum {
   unknown = -1,
   none = 0,
===================================================================
diff --git a/src/glass/src/app/shared/services/api/bootstrap.service.ts b/src/glass/src/app/shared/services/api/bootstrap.service.ts
--- a/src/glass/src/app/shared/services/api/bootstrap.service.ts	(revision 91e69960435cf1da7360f526f5829f9c3b3573ab)
+++ b/src/glass/src/app/shared/services/api/bootstrap.service.ts	(date 1614599263445)
@@ -6,8 +6,16 @@
   success: boolean;
 };

+// eslint-disable-next-line no-shadow
+export enum BootstrapStageEnum {
+  none = 0,
+  running = 1,
+  done = 2,
+  error = 3
+}
+
 export type BootstrapStatusReply = {
-  stage: 'none' | 'running' | 'done' | 'error';
+  stage: BootstrapStageEnum;
 };

 @Injectable({
```

sebastian-philipp:
Two thoughts:

1. This sounds like it should be in cephadm upstream. E.g. calling `cephadm join-cluster --token abcdef --host bootstraphost` instead of running [raw commands ](https://docs.ceph.com/en/latest/cephadm/host-management/#adding-hosts).
  And the token should be created by mgr/cephadm. I know cephadm isn't there yet unfortunately. 

2. websockets introduce a whole new level of complexity and maintenance burden. Please be super careful.

jhesketh:
>     3. Currently missing tests because I'm not fully satisfied with the direction, and don't want to be writing and rewriting tests until I'm sure the architecture makes sense.

Is this ready for review?

jecluis:
> Is this ready for review?

It is. I just can't promise it won't change direction based on feedback :)

jecluis:
@votdev changes applied to glass patch. Mind taking a look at the python bits as well?

mgfritch (approved):
generally agree with the comments made in https://github.com/aquarist-labs/aquarium/pull/225#issuecomment-787912467, but I don't think we are there yet.

tserong:
A couple of thoughts (sorry, I know I'm slightly late with this given it's already been merged):

* I'm generally OK with the approach here (websockets to do messaging, and using that to facilitate adding new nodes) especially given it moves things along, and gives us something more to play with.
* I agree with Sebastian that it would be ideal if we had this neat joining workflow in cephadm, but given we don't, taking this means we have the ability to experiment with that workflow _now_, and see how it goes. We can always try to implement this in cephadm later, or in tandem, then drop what we're doing here in favour of what ultimately becomes available in cephadm.
* Even though I said I'm OK with the approach here, I'm still slightly nervous about adding a general purpose messaging thingy. By my nervousness shouldn't stop us experimenting with it :-)

PR #226, opened 2021-03-01
gravel/glass: mark deployment ready once we finish deploying services

### LINKED ISSUE #205: gravel: /api/status does not report 'ready' stage
After bootstrapping and deploying a CephFS service the `/api/status` endpoint always returns `bootstrapped` in `deployment_state.stage`. At some point the `ready` stage is expected to be returned.

jecluis:
Once we finish deploying services, ensure we mark the deployment as finished.

Resolves #205 

### DIFF
--- a/src/gravel/controllers/bootstrap.py
+++ b/src/gravel/controllers/bootstrap.py
@@ -77,6 +77,15 @@ class Bootstrap:
 
         return True
 
+    async def mark_finished(self) -> bool:
+        stage: DeploymentStage = gstate.config.deployment_state.stage
+        if stage == DeploymentStage.ready:
+            return True
+        elif stage != DeploymentStage.bootstrapped:
+            return False
+        gstate.config.set_deployment_stage(DeploymentStage.ready)
+        return True
+
     async def get_stage(self) -> BootstrapStage:
         return self.stage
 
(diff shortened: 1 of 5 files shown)


src/gravel/controllers/bootstrap.py
```diff
@@ -77,6 +77,15 @@ async def bootstrap(self) -> bool:
 
         return True
 
+    async def mark_finished(self) -> bool:
+        stage: DeploymentStage = gstate.config.deployment_state.stage
+        if stage == DeploymentStage.ready:
+            return True
+        elif stage != DeploymentStage.bootstrapped:
+            return False
```
mgfritch, line 85: it feels as if `DeploymentStage.ready` really is the same as `DeploymentStage.bootstrapped` ...

src/gravel/controllers/bootstrap.py
```diff
@@ -77,6 +77,15 @@ async def bootstrap(self) -> bool:
...
+        gstate.config.set_deployment_stage(DeploymentStage.ready)
```
mgfritch, line 86: is it really necessary to have the web client set the ready flag? ... maybe it should be set at the end of a successful bootstrap?

src/gravel/controllers/bootstrap.py
jecluis: it isn't though - bootstrapped means we can now start deploying services, whereas 'ready' means we have gone through the whole installer.

src/gravel/controllers/bootstrap.py
jecluis: and then we have no idea if we have already went through the whole installer. :shrug:

jhesketh: Maybe we need to separate the concept of deployment_stage from "wizard/installer stage"?

jecluis:
@votdev can you check if this makes sense to you?

votdev (approved):

PR #232, opened 2021-03-01
gravel: cephfs create: ignore output from mgr call

### LINKED ISSUE #230: gravel: Can't create CephFS
**Describe the bug**
Can't create CephFS which had worked last Tuesday. Now I get first few tries an error 500 Internal Server Error than a 428 Precondition Required.

Running with "./run_aquarium.sh -d --with-system-deps" I got not more input than without debugging only:
`INFO:     192.168.121.193:34428 - "POST /api/services/create HTTP/1.1" 500 Internal Server Error`
or
`INFO:     192.168.121.193:34780 - "POST /api/services/create HTTP/1.1" 428 Precondition Required`

**To Reproduce**
Have a new cluster and try to create a new CephFS service.

**Expected behavior**
Create a working CephFS service.

jecluis:
It seems the behavior of 'fs volume create' changed beneath us and no
longer outputs any information. Ignore any possible output, and instead
rely solely on its return code.

Resolves #230

### DIFF
--- a/src/gravel/controllers/orch/cephfs.py
+++ b/src/gravel/controllers/orch/cephfs.py
@@ -33,14 +33,11 @@ class CephFS:
             "name": name
         }
         try:
-            res = self.mgr.call(cmd)
+            # this is expected to be a silent command
+            self.mgr.call(cmd)
         except CephCommandError as e:
             raise CephFSError(e) from e
-        # this command does not support json at this time, and will output
-        # free-form text instead. We are not going to parse it, but we'll make
-        # sure we've got something out of it.
-        assert "result" in res
-        assert len(res["result"]) > 0
+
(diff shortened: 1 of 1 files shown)


Devp00l (commented):
It worked for the first creation :D but not for the second.. For the second creation I got "428 Precondition Required".

sebastian-philipp:
https://github.com/ceph/ceph/pull/39147

jecluis:
> It worked for the first creation :D but not for the second.. For the second creation I got "428 Precondition Required".

The real bug here is not exposing the reasons why these things are failing (will open issues and work on different PRs), but the 428 is not a bug. It's actually a result from having too many PGs being created in the cluster.

Error status propagation needs to be clearer and better.

Devp00l (approved):

PR #247, opened 2021-03-04
doc: prefer Fixes: over Closes: in commit messages

tserong:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,8 +78,7 @@ For example:
    
    This commit introduces the first README doc for aquarist-labs.
    
-   Closes #1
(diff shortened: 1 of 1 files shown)


asettle:
Oof, controversial. I'm a "closes" gal, myself. But I don't see why not...

votdev (approved):

s0nea (approved):

tserong:
Heh...  I don't think I even knew "Closes" was a thing until I read the original CONTRIBUTING.md :-)  I switched it to Fixes because I'm used to that convention from the Ceph project, and also Fixes has already been used on a handful of commits in Aquarium.

PR #248, opened 2021-03-04
tools/build-image.sh: Use sudo with --clean

tserong:
Everything in images/build/$NAME/_out is owned by root:root, which means normal users can't remove that stuff :-/  I've split the
directory existence check so that in case the `sudo rm` fails (wrong password or whatever), it'll check again for the directory and bail out if it exists.

### DIFF
--- a/tools/build-image.sh
+++ b/tools/build-image.sh
@@ -92,13 +92,15 @@ fi
 
 build=${imgdir}/build/${build_name}
 
+if [[ -e "${build}" && "${clean}" -eq "1" ]]; then
+  echo "warning: removing existing build '${build_name}'"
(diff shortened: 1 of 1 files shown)


kshtsk:
jenkins test build image

kshtsk (dismissed):
That was in my mind too. However, I was hoping we can make it to use sudoless working directory for building image.

votdev (approved):
LGTM

tserong:
> That was in my mind too. However, I was hoping we can make it to use sudoless working directory for building image.

I wondered about that too, but didn't come up with a solution. I've added a TODO comment in the script mentioning this :-)

ml8mr (approved):
LTGM

PR #249, opened 2021-03-04
tools/build-image.sh: use /sbin/mkfs.btrfs instead of sudo

tserong:
We don't need to be root to check if mkfs.btrfs is present.

### DIFF
--- a/tools/build-image.sh
+++ b/tools/build-image.sh
@@ -85,7 +85,7 @@ if ! kiwi-ng --version &>/dev/null ; then
   error_exit "missing kiwi-ng"
 fi
 
-if ! sudo mkfs.btrfs --version &>/dev/null ; then
+if ! /sbin/mkfs.btrfs --version &>/dev/null ; then
   echo "error: missing btrfstools"
   exit 1
 fi


kshtsk:
jenkins run build-image

tools/build-image.sh
```diff
@@ -85,7 +85,7 @@ if ! kiwi-ng --version &>/dev/null ; then
   error_exit "missing kiwi-ng"
 fi
 
-if ! sudo mkfs.btrfs --version &>/dev/null ; then
+if ! /sbin/mkfs.btrfs --version &>/dev/null ; then
   echo "error: missing btrfstools"
```
kshtsk, line 89: nit: It was a typo when I committed this patch earlier, won't you mind to change 'btrfstools' to 'btrfsprogs package' too?

tools/build-image.sh
```diff
@@ -85,7 +85,7 @@ if ! kiwi-ng --version &>/dev/null ; then
   error_exit "missing kiwi-ng"
 fi
 
-if ! sudo mkfs.btrfs --version &>/dev/null ; then
+if ! /sbin/mkfs.btrfs --version &>/dev/null ; then
```
kshtsk, line 88: nice catch

kshtsk (approved):

tools/build-image.sh
votdev, line 88: Is an absolute path really necessary here?

kshtsk: @votdev that's the thing, because it is in `/sbin`:
```
> mkfs.btrfs --version ; echo $?
Absolute path to 'mkfs.btrfs' is '/usr/sbin/mkfs.btrfs', so running it may require superuser privileges (eg. root).
127
```

PR #251, opened 2021-03-04
tools/build-image.sh: s/btrfstools/btrfsprogs

tserong:


tserong:
(This is just a leftover minor tweak from #249)

kshtsk (approved):

PR #252, opened 2021-03-04
Refactor tests

jhesketh:
A series of commits to tidy up and refactor tests.

This primarily moves us to use pytest and fixtures rather than the python unittest framework.

Still a WIP because it breaks mypy and how to properly test gstate isn't obvious yet.


mgfritch (commented):
big :+1: for the fixtures!

a conftest.py that can be used by all tests is good, but I do wonder if we should leave the test_*.py files in their respective modules to avoid any potential naming conflicts?

jhesketh:
> a conftest.py that can be used by all tests is good, but I do wonder if we should leave the test_*.py files in their respective modules to avoid any potential naming conflicts?

Not sure what conflicts we might have?

We can use multiple `test_*.py` files within the `tests/` directory, and we can also create a sub-dirs if necessary. eg: `tests/test_cephadm/test_main.py`.

Sub-dirs are also how pytest scopes conftest.py inheritance (walking up the directory path) if we ever want to do some complicated division of fixtures etc.

mgfritch:
the fixtures seem shared between all components, but I'm thinking it might be worthwhile to at least split tests under the `api` and `controllers` dir?

seems as if we could easily end up with duplicate `orch.py` or `ceph.py` files in either of these modules ...

jhesketh:
> the fixtures seem shared between all components, but I'm thinking it might be worthwhile to at least split tests under the `api` and `controllers` dir?
> 
> seems as if we could easily end up with duplicate `orch.py` or `ceph.py` files in either of these modules ...

Ah, if we're creating a 1:1 map of `test_$module.py` to $module, then yes, that makes sense. I don't mind how we split them up, but I do think it's better to keep them under the one parent `tests/` folder.

jhesketh:
@mgfritch I incorporated your fixes - thanks.

The last commit in this branch also reorganises things to match the module tree. I'm not a fan of this, but it is a compromise we discussed in our meeting. If this doesn't look good to others than we can go back a commit (basically just keeping tests in arbitrary files in `tests/`), or we can place test_* files throughout the directory structure as they were before.

mgfritch (approved):
lgtm!

PR #253, opened 2021-03-04
glass: Change button text

votdev:
Change text from 'Choose selected devices' to 'Deploy devices' because it matches better what happens here.

### DIFF
--- a/src/glass/src/app/pages/deployment-page/deployment-page.component.html
+++ b/src/glass/src/app/pages/deployment-page/deployment-page.component.html
@@ -23,7 +23,7 @@
             <button mat-button
                     class="mat-stepper-next"
                     (click)="chooseDevices()">
-              <span translate>Choose selected devices</span>
+              <span translate>Deploy devices</span>
             </button>
           </div>
         </mat-step>
--- a/src/glass/src/app/pages/deployment-page/deployment-page.component.ts
+++ b/src/glass/src/app/pages/deployment-page/deployment-page.component.ts
@@ -83,7 +83,7 @@ export class DeploymentPageComponent implements OnInit {
       data: {
         type: 'yesNo',
         icon: 'warn',
-        title: TEXT('Choose selected devices'),
+        title: TEXT('Deploy devices'),
         message: TEXT(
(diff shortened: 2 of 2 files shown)


s0nea (dismissed):

ml8mr:
@asettle might oppose this, but in American English, utilize is spelled with a 'z'. Rather than starting a 'utilise' -vs- 'utilize' debate, maybe we should just use one of "(Consume|Use|Deploy) devices"? I'm even okay with "Use selected devices" (which seems better than "Choose selected devices").

jhesketh (approved):
Change looks good. But to add to the options, what about `Use selected devices`

votdev:
> @asettle might oppose this, but in American English, utilize is spelled with a 'z'. Rather than starting a 'utilise' -vs- 'utilize' debate, maybe we should just use one of "(Consume|Use|Deploy) devices"? I'm even okay with "Use selected devices" (which seems better than "Choose selected devices").

Then i'm in favor of `Deploy devices` because `(Use|Choose|...) selected devices` is incorrect here since the user can't select any device in the list. The intention of this list is to show the user which devices are detected on the node and assimilated for storage.

We discussed using `Next` as an alternative, but rejected that because we wanted to explain what happens when the button is pressed.

s0nea:
> Then i'm in favor of `Deploy devices` because `(Use|Choose|...) selected devices` is incorrect here since the user can't select any device in the list. The intention of this list is to show the user which devices are detected on the node and assimilated for storage.

I fully agree. I would be confused by the word `selected` as I actually can't select anything.

> 
> We discussed using `Next` as an alternative, but rejected that because we wanted to explain what happens when the button is pressed.

:+1:

asettle:
> @asettle might oppose this, but in American English, utilize is spelled with a 'z'. Rather than starting a 'utilise' -vs- 'utilize' debate, maybe we should just use one of "(Consume|Use|Deploy) devices"? I'm even okay with "Use selected devices" (which seems better than "Choose selected devices").

As much as I fundamentally disagree, EN-US is the standard. It needs to be consistently used across our dashboard and doc.

s0nea (approved):

PR #256, opened 2021-03-05
test/vagrant: Always add new Vagrant box

tserong:
The first time you run `tests/vagrant/setup.sh`, it adds a new Vagrant box.  That's fine.  But if you later make some changes and build a new image, then run `tests/vagrant/setup.sh` *again*, you end up using the old box, because something with that name already exists.  Let's delete that thing if it exists, so we're always using the latest and greatest.

### DIFF
--- a/tests/vagrant/setup.sh
+++ b/tests/vagrant/setup.sh
@@ -64,33 +64,32 @@ done
   echo "error: setup with name '${setup_name} already exists" && \
   exit 1
 
-echo "=> Searching for Vagrant box '${box}' ..."
-if ! vagrant box list | grep -q "^${box}[ ]\+" ; then
-  echo "=> Vagrant box '${box}' not found, searching for existing build"
-
-  imgdir=$(realpath ${basedir}/../../images)
-
-  [[ ! -e "${imgdir}" ]] && \
-    echo "error: unable to find images directory" && \
-    exit 1
-
-  [[ ! -e "${imgdir}/build" ]] && \
-    echo "error: unable to find images build directory" && \
-    exit 1
-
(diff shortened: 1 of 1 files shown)


tserong:
BTW, this change is much less horrible if you ignore whitespace:

```
> git show -w
[...]
diff --git a/tests/vagrant/setup.sh b/tests/vagrant/setup.sh
index 44e5ec1..68698cb 100755
--- a/tests/vagrant/setup.sh
+++ b/tests/vagrant/setup.sh
@@ -64,9 +64,10 @@ done
   echo "error: setup with name '${setup_name} already exists" && \
   exit 1
 
-echo "=> Searching for Vagrant box '${box}' ..."
-if ! vagrant box list | grep -q "^${box}[ ]\+" ; then
-  echo "=> Vagrant box '${box}' not found, searching for existing build"
+if vagrant box list | grep -q "^${box}[ ]\+" ; then
+  echo "=> Removing existing Vagrant box '${box}'"
+  vagrant box remove ${box}
+fi
 
 imgdir=$(realpath ${basedir}/../../images)
 
@@ -87,10 +88,8 @@ if ! vagrant box list | grep -q "^${box}[ ]\+" ; then
   echo "error: unable to find vagrant libvirt image for '${box}'" && \
   exit 1
 
+echo "=> Adding new Vagrant box '${box}'"
 vagrant box add ${box} ${img} || exit 1
-else
-  echo "=> Vagrant box '${box}' already exists"
-fi
 
 sdir=${setupdir}/${setup_name}
 mkdir -p ${sdir} || exit 1
```

jecluis (approved):

PR #260, opened 2021-03-05
Adding Debian10 support for the setup script

AvengerMoJo:
Allow the setup script to run in Debian10 environment.


jecluis (approved):
I'm not testing this, so I'm assuming it works as it seems.

PR #261, opened 2021-03-08
glass: Set the bootstrap message

votdev:


votdev:
Suggestions and improvements are welcome.

Devp00l (dismissed):

s0nea (dismissed):

src/glass/src/app/pages/bootstrap-page/bootstrap-page.component.html
```diff
@@ -13,7 +13,7 @@
         <div fxLayout="column"
              fxLayoutAlign="center center">
           <p>
-            <span translate>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Proin porttitor quam convallis hendrerit lacinia. Integer consectetur metus eu nibh fringilla, ac lobortis nisl porttitor. Pellentesque habitant morbi tristique senectus et netus et malesuada fames ac turpis egestas. Donec at purus odio.</span>
+            <span translate>Aquarium will now install and setup the core system. All recognized storage devices are automatically assigned to the data pool. Afterwards you will be able to configure the services you want to provide.</span>
```
jecluis, line 16: I'd drop the second phrase. We don't have a single data pool, nor do we assign the devices to any pool whatsoever. While _I_ understand what you mean, I don't think it would be obvious to the user. :)

src/glass/src/assets/i18n/aquarium.pot
```diff
@@ -13,9 +13,18 @@ msgstr ""
...
 
+msgid "Address"
+msgstr ""
+
 msgid "Aquarium"
 msgstr ""
 
+msgid ""
+"Aquarium will now install and setup the core system. All recognized storage "
+"devices are automatically assigned to the data pool. Afterwards you will be "
```
jecluis, line 24: same here

jecluis (changes requested):
I don't think we should be talking about "data pool" in this message, and especially not at this stage.

src/glass/src/app/pages/bootstrap-page/bootstrap-page.component.html
asettle, line 16:
```suggestion
            <span translate>Aquarium will now install and setup the core system. All recognized storage devices are automatically assigned. Afterwards, you will be able to configure the services you want to provide.</span>

asettle: Just a suggestion to keep it vague

asettle (commented):
Just a suggestion inline, otherwise, looks good

src/glass/src/app/pages/bootstrap-page/bootstrap-page.component.html
jecluis: Yeah, that looks better IMO

jecluis (approved):

Issue #32, opened 2021-02-02
gravel: persistent state


jecluis:
Ideally distributed from the start, with failover, etc.

Would be cool not having to reinvent the wheel.

I, personally, don't think relying on ceph-mon's config-key store is the right way to go.

theanalyst:
How are we going for persistence here, a kv-store in hte likes of etcd/consul and the like or going native like rocksdb?

tserong:
A thought that came up in one of our calls is to use ceph-mon's config store, but have a local cache of any details we need, on each aquarium node. So most of the time (when ceph is actually running) all aquarium nodes have access to everything. If the cluster is down (or, at least, the mons aren't in quorum), any individual aquarium node would fall back to using its local cache, until the cluster comes back. That means aquarium can still try to do something useful in the failure case, and also means we don't need a separate cluster aware / replicated config store.

PR #266, opened 2021-03-09
gravel: add support to extend cluster to multiple nodes

### LINKED ISSUE #34: gravel: extend to two nodes
This requires some thought and research: what is the best approach to achieve this?

jecluis:
With this patchset (that devolved into a little bit more than just node extension), we are now able to support adding a new node to the cluster.

Essentially, provided the user has specified the leader's address and a valid token, the node wanting to join will send a message to the leader stating its intentions. The leader will then either error out, in case of invalid token or missing parameters, or acknowledge the intent by sending a welcome message with the cluster's public key. Once this key has been written to disk, the node will flag the leader as being ready to add, and the leader will then invoke `orch host add`.

This is a fairly straightforward workflow, but several turns had to be taken until we reached here.

This patchset is currently breaking the tests, and I haven't yet had an opportunity to figure out why.

Resolves #34 


src/gravel/controllers/nodes/mgr.py
```diff
@@ -138,14 +138,14 @@ def __init__(self):
         logger.debug(f"=> mgr -- init > {self._state}")
 
         assert self._state.stage == NodeStageEnum.NONE or \
-               self._state.stage == NodeStageEnum.BOOTSTRAPPED or \
-               self._state.stage == NodeStageEnum.READY
+            self._state.stage == NodeStageEnum.BOOTSTRAPPED or \
+            self._state.stage == NodeStageEnum.READY
```
jhesketh, line 142: (nit) You could use `assert self._state.stage in [...]`

src/gravel/controllers/nodes.py
```diff
@@ -171,8 +172,21 @@ async def join(self, address: str, token: str) -> bool:
...
         await conn.send(msg)
+
         reply: MessageModel = await conn.receive()
         logger.debug(f"=> mgr -- join > recv: {reply}")
+        assert reply.type == MessageTypeEnum.WELCOME
+        welcome = WelcomeMessageModel.parse_obj(reply.data)
+        assert welcome.cluster_uuid
+        assert welcome.pubkey
+
+        authorized_keys: Path = Path("/root/.ssh/authorized_keys")
```
jhesketh: I don't like this hardcoded authorized keys, or the fact that aquarium will modify the system keys.

However I think the alternatives in the way they would need to interoperate with cephadm are too complex to worry about for now.

src/gravel/controllers/nodes.py
```diff
@@ -163,7 +171,17 @@ def __init__(self):
         self._incoming_task = asyncio.create_task(self._incoming_msg_task())
 
     async def join(self, address: str, token: str) -> bool:
-        logger.debug(f"=> mgr -- join > with addr {address}, token: {token}")
+        logger.debug(f"=> mgr -- join > with leader {address}, token: {token}")
+
+        if self._state:
+            assert self._state.role == NodeRoleEnum.LEADER or \
+                   self._state.role == NodeRoleEnum.FOLLOWER
```
jhesketh: (nit) wouldn't it be easier and more future proof to check that `self._state.role == NodeRoleEnum.none`?

src/gravel/controllers/nodes.py
```diff
@@ -160,13 +169,22 @@ class NodeMgr:
...
     def __init__(self):
-        self._connmgr = ConnMgr()
         self._shutting_down = False
+        self._connmgr = ConnMgr()
+        self._manifest = None
+        self._token = None
+
+        self._init_node()
+        assert self._state
+        if self._state.stage == NodeStageEnum.READY:
```
jhesketh: if the state isn't ready at this point, the NodeMgr is never started. Should this wait on an event to change the node state?

Shouldn't this also use the `self.ready` property?

src/gravel/controllers/bootstrap.py
```diff
@@ -50,11 +55,16 @@ async def _is_bootstrapping(self) -> bool:
     async def bootstrap(self) -> bool:
         logger.debug("bootstrap > do bootstrap")
 
-        if not await self._should_bootstrap():
```
jhesketh, line 53: Is `_should_bootstrap` redundant now and needs removing?

src/gravel/controllers/nodes.py
```diff
@@ -225,29 +237,40 @@ async def join(self, address: str, token: str) -> bool:
 
         return True
 
+    async def prepare_bootstrap(self) -> None:
+        assert self._state
+        if self._state.stage > NodeStageEnum.NONE:
```
jhesketh: (nit) maybe `!=`?

src/gravel/controllers/nodes.py
```diff
@@ -200,32 +228,45 @@ def start(self):
...
-        if deployment_stage != DeploymentStage.none:
+        assert self._state
+        if self._state.stage == NodeStageEnum.BOOTSTRAPPING:
+            raise NodeBootstrappingError()
+        elif self._state.stage == NodeStageEnum.BOOTSTRAPPED:
             raise NodeHasBeenDeployedError()
+        elif self._state.stage == NodeStageEnum.JOINING:
+            raise NodeAlreadyJoiningError()
+        elif self._state.stage == NodeStageEnum.READY:
+            raise NodeHasJoinedError()
```
jhesketh: (nit) this set of checks could be simplified.. eg check if it's not in desired state and raise an unexpected state error that logs current state.

jhesketh (changes requested):
Haven't managed to make it through my full review. Most of my comments are nits and I like the general direction this is heading.

I've reviewed the commits up to and including `6c13e9a674f055fbb7c2c2a7e5a0d472849765ae` (` gravel: don't track deployment state in config`)

src/gravel/controllers/bootstrap.py
jecluis: yeah, missed that. thanks.

src/gravel/controllers/nodes.py
jecluis: this code is somewhat obsoleted by a following commit at some stage of the patchset. Sorry :\

src/gravel/controllers/nodes.py
jecluis: This is where cephadm is dropping the key during bootstrap, and I could not think of any other (reasonable) way to achieve this. We need that public key to be in the node's authorized_keys so the orchestrator can reach out to the node.

As for changing system keys, I mean... We are an opinionated appliance. We ought to be able to do opinionated things :)

src/gravel/controllers/config.py
```diff
@@ -62,12 +60,8 @@ def __init__(self, path: str = config_dir):
 
         if not self.confpath.exists():
             initconf: ConfigModel = ConfigModel(
-                version=3,
-                name="",
-                deployment_state=DeploymentStateModel(
-                    last_modified=datetime.now(),
-                    stage=DeploymentStage.none
-                )
+                version=1,
```
tserong, line 63: Did that version number just go backwards?

tserong (commented):
Just a couple of comments so far (see inline).

Also, I'm wondering about NodeRoleEnum...  Maybe I'm reading this wrong, but it looks like the initial bootstrap node is given the leader role, and everything else is a follower? If that's the case, then you can only join a cluster by pointing a new node to the original bootstrap node. Shouldn't it be the case that you can join a cluster by pointing a new node at _any_ existing node, not just the first/bootstrap node?

src/gravel/controllers/nodes.py
```diff
@@ -171,8 +172,21 @@ async def join(self, address: str, token: str) -> bool:
...
+        welcome = WelcomeMessageModel.parse_obj(reply.data)
+        assert welcome.cluster_uuid
+        assert welcome.pubkey
+
+        authorized_keys: Path = Path("/root/.ssh/authorized_keys")
+        if not authorized_keys.parent.exists():
+            authorized_keys.parent.mkdir(0o700)
+        with authorized_keys.open("a") as fd:
+            fd.writelines([welcome.pubkey])
+            logger.debug(f"=> mgr -- join > wrote pubkey to {authorized_keys}")
```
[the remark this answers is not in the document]
jecluis: yeah, at the moment that is very much a possibility.

Currently we are not keeping track of existing nodes. That ought to be done in the manifest, but seemed like adding more stuff on top of this pile would not help.

Essentially, we'll be adding a node to the manifest just as we add call on `orch host add`, and then we will not only persist the manifest including the new node, but ideally share it with other nodes. That way we can prevent a given node from joining multiple times. We'll likely want to do that by hostname/address; or using the node's uuid, ultimately, if that is not enough.

src/gravel/controllers/config.py
tserong: OK, I'll join you in that shrug :-)

tserong:
> So, yeah, clusters all the way down.

Dammit. I would love for this thing to be leaderless (and followerless). But I think that's probably a separate question/discussion and needn't hold up this PR.

src/gravel/controllers/nodes.py
jhesketh: > This is where cephadm is dropping the key during bootstrap, and I could not think of any other (reasonable) way to achieve this. We need that public key to be in the node's authorized_keys so the orchestrator can reach out to the node.

Right, that's why I mentioned it'd be difficult to do differently.

> As for changing system keys, I mean... We are an opinionated appliance. We ought to be able to do opinionated things :)

But yes, very good point. :+1:

jhesketh:
>     1. We need to keep track of which nodes are in the cluster, so we prevent a node from joining multiple times, and that should be by relying on a manifest, which ought to be shared across all nodes;

Does using something like etcd for things like the manifest allow us to head towards leaderless more readily (if not immediately)?

tserong:
> Does using something like etcd for things like the manifest allow us to head towards leaderless more readily (if not immediately)?

Do we even *need* a manifest? (And can we carry on that discussion either #268 or #32, as I think that's a bigger question)

src/gravel/controllers/nodes/conn.py
```diff
@@ -55,18 +65,29 @@ class ConnMgr:
...
         self._conns = []
         self._passive_conns = []
         self._active_conns = []
         self._conn_by_endpoint = {}
         self._incoming_queue = asyncio.Queue()
 
+    def start(self) -> None:
+        self._is_started = True
+
+    def is_started(self) -> bool:
```
jhesketh, line 56: (nit) should be `@property` and maybe just `started`?

src/gravel/api/orch.py
```diff
@@ -113,7 +113,7 @@ async def get_node_info() -> NodeInfoModel:
 
 @router.get("/inventory", response_model=NodeInfoModel)
 async def get_inventory() -> NodeInfoModel:
-    latest = await inventory.latest()
+    latest = inventory.get_inventory().latest
```
jhesketh, line 116: doesn't this need to be `latest()`?

src/gravel/controllers/resources/storage.py
```diff
@@ -127,3 +127,11 @@ async def _update(self) -> None:
...
         self._state.pools_by_name = by_name
         self._state.pools_by_id = by_id
+
+
+_storage = Storage()
+
+
+def get_storage() -> Storage:
+    global _storage
+    return _storage
```
jhesketh, line 137: Really these are still declared at a module state.. (referring to _inventory too).

I don't think we need to block this PR on this, but I think we should reconsider how we are keeping and passing state/context around.

src/gravel/controllers/resources/inventory.py
```diff
@@ -37,10 +50,26 @@ async def probe(self) -> None:
         diff: int = int(time.monotonic()) - start
         logger.info(f"=> inventory probing took {diff} seconds")
         self._latest = nodeinfo
+        await self._publish()
 
-    async def latest(self) -> Optional[NodeInfoModel]:
+    @property
+    def latest(self) -> Optional[NodeInfoModel]:
```
jhesketh, line 56: This will fix my comment on the previous commit. (Sorry just going through this in commit order).

src/gravel/controllers/resources/inventory.py
```diff
@@ -37,10 +50,26 @@ async def probe(self) -> None:
...
 
+    def subscribe(
+        self,
+        cb: Callable[[NodeInfoModel], Awaitable[None]],
+        once: bool
+    ) -> None:
+        self._subscribers.append(Subscriber(cb=cb, once=once))
+
+    async def _publish(self) -> None:
+        assert self._latest
```
jhesketh, line 67: Probably should use self.latest?

jhesketh (dismissed):
Overall I am okay with this PR but do have some architectural directions that I think we should discuss. Primarily I share concerns about orchestrators all the way down, and also the handling of state/scope. The latter we can address as a followup (if necessary/possible), and the former is more something to note.

src/gravel/controllers/resources/storage.py
jecluis: they are, but they are no longer at the `resources/__init__.py` level, which cause a whole lot of different problems during imports.

And, aside from having everything being initiated on `GlobalState()` and then having only GlobalState as a single global state class, I don't see how we can have these things initialized and ready to go to be used in the fastapi routes.

Unfortunately, the way these are created makes it  impossible (at least I haven't found an alternative) to pass them context to be handled on a file-by-file/class-by-class basis.

src/gravel/controllers/resources/inventory.py
jecluis: Personally I'm a fan of keeping assertions on internal state separate from properties. As I see it, properties are for the benefit of consumers; in this case we are returning `self._latest` on the property, as is, but we could, technically, be doing other foo that could, possibly, not be applicable to this assertion.

jhesketh: Depends if we expect the method providing the property to change. However, this is an assert so that's reasonable to me.

src/gravel/controllers/resources/storage.py
jhesketh: Right, having the GlobalState handle everything is something I think that is worth considering.

jecluis: Lets attempt that in the near future then :)

Just filed https://github.com/aquarist-labs/aquarium/issues/276 to track that.

src/gravel/tests/unit/controllers/test_config.py
```diff
@@ -1,19 +1,11 @@
...
-    import Config, DeploymentStage
+from gravel.controllers.config import Config
 
 
 def test_config_version(fs):
     config = Config()
-    assert config.config.version == 3
-
-
-def test_deployment_state(fs):
```
jhesketh, line 13: We need to open a ticket to re-add this and the disabled_test_services if this gets merged.

jhesketh (approved):

src/gravel/tests/unit/controllers/test_config.py
jecluis: this was removed because `deployment_state` no longer exists. We might want to test `node_state`, but unless it's through a functional test I'm not sure it's really worth it.

PR #267, opened 2021-03-09
Adding Ubuntu20.10 setup script support

AvengerMoJo:
Adding Ubuntu20.10 support 
 
1) Replacing the sed with the os-release file may or may not have quote after ID=
2) swap the order by installing the repo key before adding repo to avoid add-repo update without key. 


AvengerMoJo:
Thanks I have tested in a fresh download install Ubuntu20.10

PR #265, opened 2021-03-08
add config for testing the PR using wasser

kshtsk:
Signed-off-by: Kyr Shatskyy <kyrylo.shatskyy@suse.com>

PR #269, opened 2021-03-09
microOS: build other image types besides Vagrant's

jecluis:

### DIFF
--- a/tools/build-image.sh
+++ b/tools/build-image.sh
@@ -25,7 +25,12 @@ usage: $0 [options]
 options:
   -n NAME | --name NAME     Specify build name (default: aquarium).
   -c | --clean              Cleanup an existing build directory before building.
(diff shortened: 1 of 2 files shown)


kshtsk:
@jecluis this looks good, though I've added python3-websockets related fix within #265, if you like to merge your PR earlier, I gonna drop it from mine.

tools/build-image.sh
```diff
@@ -81,6 +90,21 @@ done
 [[ -z "${build_name}" ]] && \
   usage_error_exit "missing build name"
 
+[[ -z "${imgtype}" ]] && \
+  usage_error_exit "image type must be provided"
```
kshtsk, line 94: nit: not sure this is really necessary since there is a default value for _imgtype_.

kshtsk (approved):

PR #272, opened 2021-03-09
Add kiwi boxbuild for distro other than openSUSE

AvengerMoJo:
Adding python3-kiwi-boxed-plugin so distribution other than openSUSE could use boxbuild to create an image without adjusting the image creation process. Meaning other distributions will be downloading an openSUSE build env to create microOS, instead of building the image with native env. It provides a cross distributions build script without modifying the kiwi config one by one.
The downside of using boxbuild in other distribution is that requires downloading around 600MB of build environment and the kiwi build will be running in a qemu environment. 

Update the build script for Debian and Ubuntu to use the boxbuild option with kiwi-ng 

### DIFF
--- a/tools/build-image.sh
+++ b/tools/build-image.sh
@@ -188,11 +188,27 @@ cp ${imgdir}/microOS/config.{sh,xml} ${build}/
 bundle || exit 1
 
 mkdir ${build}/{_out,_logs}
-(set -o pipefail
-sudo kiwi-ng --debug --profile=${profile} --type oem \
-  system build --description ${build} \
-  --target-dir ${build}/_out |\
-  tee ${build}/_logs/${build_name}-build.log)
+
+osid=$(grep '^ID=' /etc/os-release | sed -e 's/\(ID=["]*\)\(.\+\)/\2/' | tr -d '"')
+case $osid in
+  opensuse-tumbleweed | opensuse-leap)
+    (set -o pipefail
+    sudo kiwi-ng --debug --profile=${profile} --type oem \
+      system build --description ${build} \
+      --target-dir ${build}/_out |\
+      tee ${build}/_logs/${build_name}-build.log)
+    ;;
+  debian | ubuntu)
+    (set -o pipefail
+    sudo kiwi-ng --debug --profile=${profile} --type oem \
+      system boxbuild --box tumbleweed --no-update-check -- --description ${build} \
+      --target-dir ${build}/_out |\
+      tee ${build}/_logs/${build_name}-build.log)
+    ;;
+  *)
+    echo "error: unsupported distribution ($osid) kiwi-ng may not work"
+    exit 1
+      ;;
+esac
 
 exit $?
-
--- a/tools/setup-dev.sh
+++ b/tools/setup-dev.sh
@@ -11,6 +11,7 @@ dependencies_opensuse_tumbleweed=(
   "npm"
   "vagrant"
   "vagrant-libvirt"
+  "libvirt-daemon"
 )
 
 dependencies_debian=(
@@ -20,9 +21,11 @@ dependencies_debian=(
   "python3-pip"
   "python3-rados"
   "python3-kiwi"
+  "python3-kiwi-boxed-plugin"
   "python3-venv"
   "nodejs"
   "vagrant"
+  "libvirt-daemon"
 )
 
 dependencies_ubuntu=(
@@ -32,9 +35,11 @@ dependencies_ubuntu=(
   "python3-pip"
   "python3-rados"
   "python3-kiwi"
+  "python3-kiwi-boxed-plugin"
   "python3-venv"
   "nodejs"
   "vagrant"
+  "libvirt-daemon"
 )
 
 usage() {


ml8mr (commented):
I don't have a machine with nested virt handy, but did test this in a VM. OS detection works well, and the kiwi-ng command worked for that environment. (I'll try to track down a test machine and spin it up there shortly.)

ml8mr (approved):
Testing on real hardware went well, so this is approved from my side. (As a small sidenote, we may want to add libvirt components to the dependency list.)

votdev:
Hmmm, sounds like https://github.com/aquarist-labs/aquarium/tree/main/images/microOS_build_env with another approach. Can't we harmonize them?

AvengerMoJo:
Yes, the benefits of using kiwi boxbuild is eliminated our effort to support cross platform build and it allows a single kiwi config to generate the same image across different distro. 
I am unsure but maybe using docker build for different platform may need up require us to maintain more build environment work and multiple docker file for different platform.

votdev:
> Yes, the benefits of using kiwi boxbuild is eliminated our effort to support cross platform build and it allows a single kiwi config to generate the same image across different distro.
> I am unsure but maybe using docker build for different platform may end up require us to maintain more build environment work and multiple docker file for different platform.

The idea of the Docker container is to have only one build environment (based on openSUSE Leap) that can be run on every platform to build the MicroOS Vagrant box.

AvengerMoJo:
Right, I understand. That's the reason I am using the kiwi boxbuild in the same build script all platform. So we don't have to maintain multiple build script for different platform.

tools/build-image.sh
```diff
@@ -85,7 +85,9 @@ if ! kiwi-ng --version &>/dev/null ; then
   error_exit "missing kiwi-ng"
 fi
 
-if ! /sbin/mkfs.btrfs --version &>/dev/null ; then
+
+MKFS_BTRFS=`which mkfs.btrfs`
```
jecluis, line 89: Apparently, using `which` introduces a new dependency, and I've been advised to refrain from using it.

jecluis (commented):
While I enjoy the idea of having one single script, relying on kiwi, to build all images for whatever flavors, I'm stuck on figuring out why we want this. I'm guessing this is meant to enable users running on distros other than opensuse to build our microOS image, but I'm not sure what benefits this has over @votdev 's docker-based builder (which, IIRC, basically does the same thing but building on an opensuse image instead).

AvengerMoJo:
> While I enjoy the idea of having one single script, relying on kiwi, to build all images for whatever flavors, I'm stuck on figuring out why we want this. I'm guessing this is meant to enable users running on distros other than opensuse to build our microOS image, but I'm not sure what benefits this has over @votdev 's docker-based builder (which, IIRC, basically does the same thing but building on an opensuse image instead).

Understand, I'm not trying to say which is better. But my idea is for us to only need to maintain one build process and the kiwi boxbuild come free from the project. Therefore we don't have to maintain it in a long run.

jecluis:
Yeah, I don't have a strong enough opinion to say which way would be ideal. Maybe @tserong has an opinion?

tserong:
It looks like the difference is that with this PR, you can run `./tools/build-image.sh` immediately on any OS (well, on openSUSE or Debian/Ubuntu so far), whereas with the Docker container approach you first have to build and start the docker container on non-openSUSE distros, _then_ you can run `./tools/build-image.sh`. So it's a slightly different workflow. This PR means we have a simpler process, but also means we need to add other supported distros to the list in that `case` statement. The Docker container means a different process, but "should work anywhere" automatically.

Having said all that, TBH I don't really have a strong preference either way myself. Sorry!

jecluis:
The one benefit of one over the other that comes to mind is that with the kiwi approach we do cache packages, so fewer downloads between builds (which is cool if you are building an image very frequently).

Then again, given I don't have a stronger opinion than that, I'd say have them both until a time where one becomes the clear winner? :shrug:

tserong:
> Then again, given I don't have a stronger opinion than that, I'd say have them both until a time where one becomes the clear winner? shrug

Fine by me :-)

jecluis:
Okay. @AvengerMoJo let's fix the conflict and then I'll be happy to approve and merge.

tools/build-image.sh
```diff
@@ -188,11 +188,27 @@ cp ${imgdir}/microOS/config.{sh,xml} ${build}/
...
+  opensuse-tumbleweed | opensuse-leap)
+    (set -o pipefail
+    sudo kiwi-ng --debug --profile={profile} --type oem \
+      system build --description ${build} \
+      --target-dir ${build}/_out |\
+      tee ${build}/_logs/${build_name}-build.log)
+    ;;
+  debian | ubuntu)
+    (set -o pipefail
+    sudo kiwi-ng --debug --profile=Ceph-Vagrant --type oem \
```
jecluis, line 203: you might want to change this to `${profile}` as well, instead of `Ceph-Vagrant`?

jecluis (changes requested):
Also, not that it is too much of a blocker, it would be nicer to have the commit rebased on top of  main rather than a merge :shrug:

kshtsk:
Integration test failed:
```
>>> + sudo kiwi-ng --debug '--profile={profile}' --type oem system build --description /home/opensuse/aquarium/images/build/aquarium --target-dir /home/opensuse/aquarium/images/build/aquarium/_out
>>> + tee /home/opensuse/aquarium/images/build/aquarium/_logs/aquarium-build.log
>>> [ DEBUG   ]: 11:23:48 | EXEC: [mkdir -p /home/opensuse/aquarium/images/build/aquarium/_out/build]
>>> [ INFO    ]: 11:23:48 | Loading XML description
>>> [ INFO    ]: 11:23:48 | Support for XML markup available
>>> [ ERROR   ]: 11:23:49 | KiwiProfileNotFound: profile {profile} not found for host arch x86_64
```

Someone forgot a dollar sign.

tools/build-image.sh
```diff
@@ -188,11 +188,27 @@ cp ${imgdir}/microOS/config.{sh,xml} ${build}/
...
-  system build --description ${build} \
-  --target-dir ${build}/_out |\
-  tee ${build}/_logs/${build_name}-build.log)
 
-exit $?
+osid=$(grep '^ID=' /etc/os-release | sed -e 's/\(ID=["]*\)\(.\+\)/\2/' | tr -d '"')
+case $osid in
+  opensuse-tumbleweed | opensuse-leap)
+    (set -o pipefail
+    sudo kiwi-ng --debug --profile={profile} --type oem \
```
kshtsk, line 196: This is caught by jenkins job.

AvengerMoJo:
> Also, not that it is too much of a blocker, it would be nicer to have the commit rebased on top of main rather than a merge
Rebased but I am not sure how to squash it into a single commit

tools/build-image.sh
AvengerMoJo: fix that

tools/build-image.sh
AvengerMoJo: fix that

jecluis (approved):

jecluis (changes requested):
somehow we still have 4 commits on this PR. I fear this might break history. _sigh_

jecluis:
@AvengerMoJo if you don't mind, something like this should fix the branch just well enough to merge:

```
$ git checkout origin/main -b fancy-branch
$ git cherry-pick 9c9427f4f615520399117ba611c459caba5a4bcd 5812dcd7a93fa2ece791c3253aa914f92058e38e
$ git push your-fork-remote fancy-brach
```

And open a new PR with that branch. We can then squash the two commits upon merge.

And I would very much encourage not developing on your main branch, but that's just my preference :)

AvengerMoJo:
close this and replace it with #285

PR #273, opened 2021-03-10
gravel: Add option to configure port via cmd args

votdev:
Add the command line argument "--port" to run_aquarium.sh to change the port gravel is listening to. Defaults to 1337.

### DIFF
--- a/run_aquarium.sh
+++ b/run_aquarium.sh
@@ -23,12 +23,14 @@ has_config=false
 config_path=""
 is_new=false
 with_systemdeps=false
+port=1337
 
 while [[ $# -gt 0 ]]; do
   case $1 in
     -d|--debug) is_debug=true ;;
     -c|--config) has_config=true ; config_path=$2 ; shift 1 ;;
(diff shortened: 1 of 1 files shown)


run_aquarium.sh
```diff
@@ -23,12 +23,14 @@ has_config=false
...
 is_new=false
 with_systemdeps=false
+port=1337
 
 while [[ $# -gt 0 ]]; do
   case $1 in
     -d|--debug) is_debug=true ;;
     -c|--config) has_config=true ; config_path=$2 ; shift 1 ;;
     -n|--new) is_new=true ;;
+    -p|--port) port=$2 ; shift 1 ;;
```
jecluis, line 33: this needs a check for whether `${port}` is populated afterwards. Something like `./run_aquarium --port` or `./run_aquarium --port ""` would end up with an empty variable.

jecluis (changes requested):

jecluis:
@votdev will you be pushing for this still?

jecluis (approved):

PR #285, opened 2021-03-11
Add kiwi boxbuild for distro other then openSUSE

AvengerMoJo:
Replacing the old PR #272  altogether. 

### DIFF
--- a/tools/build-image.sh
+++ b/tools/build-image.sh
@@ -188,11 +188,27 @@ cp ${imgdir}/microOS/config.{sh,xml} ${build}/
 bundle || exit 1
 
 mkdir ${build}/{_out,_logs}
-(set -o pipefail
-sudo kiwi-ng --debug --profile=${profile} --type oem \
-  system build --description ${build} \
-  --target-dir ${build}/_out |\
-  tee ${build}/_logs/${build_name}-build.log)
+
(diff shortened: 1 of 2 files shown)


jecluis (dismissed):

tools/build-image.sh
```diff
@@ -188,11 +188,27 @@ cp ${imgdir}/microOS/config.{sh,xml} ${build}/
...
+      --target-dir ${build}/_out |\
+      tee ${build}/_logs/${build_name}-build.log)
+    ;;
+  *)
+    echo "error: unsupported distribution ($osid) kiwi-ng may not work"
+    exit 1
+      ;;
+esac
 
+exit $?
```
jecluis, line 214: I wonder though whether this return code is going to be for the actual command that was run, or, as bash sometimes likes to do, the last statement (which I'm guessing was a successful `case`).

As much as I hate to hold up the PR yet again, mind adding the `exit $?` right after the calls to kiwi (mimicking what had been done)? Unless you are certain this is propagating the error as expected?

AvengerMoJo: sound good, I just move the exit $? into the case switch. 
Meanwhile I updated btrfs file check with different distro location

jecluis (approved):

PR #286, opened 2021-03-11
gravel: api/services: available space must account for reservations

### LINKED ISSUE #236: gravel: reservations API not providing raw reservation, just service reservation
Essentially we're not providing the caller with the raw reserved space (i.e., service reservation * replicas), which makes the UI (and thus the user) believe they should have a lot more space available than they actually do.

jecluis:
Resolves #236

### DIFF
--- a/src/gravel/controllers/services.py
+++ b/src/gravel/controllers/services.py
@@ -11,7 +11,10 @@ from gravel.controllers.orch.cephfs import CephFS, CephFSError
 from gravel.controllers.orch.models \
     import CephFSListEntryModel, CephOSDPoolEntryModel
 from gravel.controllers.gstate import gstate
-from gravel.controllers.resources import storage
+from gravel.controllers.resources.storage import (
+    Storage,
+    get_storage
+)
 
 
 class ServiceError(Exception):
@@ -110,6 +113,12 @@ class Services:
             total += (service.reservation * service.replicas)
         return total
 
+    @property
+    def available_space(self) -> int:
+        storage: Storage = get_storage()
+        total_storage: int = storage.total
+        return (total_storage - self.total_raw_reservation)
+
     def __contains__(self, name: str) -> bool:
         return name in self._services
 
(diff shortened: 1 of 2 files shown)


mgfritch (approved):

src/gravel/controllers/services.py
```diff
@@ -110,6 +113,12 @@ def total_raw_reservation(self) -> int:
             total += (service.reservation * service.replicas)
         return total
 
+    @property
+    def available(self) -> int:
```
jhesketh, line 117: This property isn't very clearly named. The object is `Services` so in the context of thinking about services "available" does not make me think of available space.

jhesketh (commented):
Just a nit around the naming of the property. The change otherwise looks fine.

(Another thought could be to have this as part of the Node object instead)

votdev (approved):
LGTM, but requires some adaption in the UI. Do you want to do that in this PR or shall i open a glass related PR?

```
--- a/src/glass/src/app/core/dashboard/widgets/capacity-dashboard-widget/capacity-dashboard-widget.component.ts	(revision b6fca28e7bf4328bb086a292fb6308f9c4a20600)
+++ b/src/glass/src/app/core/dashboard/widgets/capacity-dashboard-widget/capacity-dashboard-widget.component.ts	(date 1615545673762)
@@ -31,8 +31,8 @@
     // @ts-ignore
     this.subscription = this.loadDataEvent.subscribe(() => {
       this.chartData = [
-        { name: translate(TEXT('Assigned')), value: this.data.available - this.data.reserved },
-        { name: translate(TEXT('Unassigned')), value: this.data.reserved }
+        { name: translate(TEXT('Assigned')), value: this.data.reserved },
+        { name: translate(TEXT('Unassigned')), value: this.data.available }
       ];
     });
   }
```

jecluis:
if you could do that in another PR I'd appreciate it immensely :)

PR #287, opened 2021-03-11
gravel: allow creating pools with one replica

### LINKED ISSUE #238: gravel: set allow setting pool size 1 on service creation
Without disabling a safety flag before service creation, creating a pool with only one replica is impossible and results in error.

jecluis:
Resolves #238 

### DIFF
--- a/src/gravel/controllers/nodes/mgr.py
+++ b/src/gravel/controllers/nodes/mgr.py
@@ -341,6 +341,12 @@ class NodeMgr:
             logger.error("=> mgr -- unable to allow pool size 1")
             logger.debug(str(e))
 
+        try:
+            mon.disable_warn_on_no_redundancy()
+        except CephCommandError as e:
+            logger.error("=> mgr -- unable to disable redundancy warning")
+            logger.debug(str(e))
+
     async def finish_deployment(self) -> None:
         assert self._state
 
--- a/src/gravel/controllers/nodes/mgr.py
+++ b/src/gravel/controllers/nodes/mgr.py
@@ -29,6 +29,10 @@ from fastapi import status
 from fastapi.logger import logger as fastapi_logger
 from gravel.cephadm.models import NodeInfoModel
 from gravel.controllers.gstate import gstate
+from gravel.controllers.orch.ceph import (
+    CephCommandError,
+    Mon
+)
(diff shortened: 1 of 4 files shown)


src/gravel/controllers/orch/ceph.py
```diff
@@ -156,10 +179,48 @@ def get_pools(self) -> List[CephOSDPoolEntryModel]:
...
+        except CephCommandError as e:
+            logger.error("=> ceph -- mon > unable to set pool size")
+            logger.debug(str(e))
+            logger.debug(str(size_cmd))
+
+        minsize = 2 if size > 2 else 1
+        minsize_cmd: Dict[str, str] = {
+            "prefix": "osd pool set",
+            "pool": name,
+            "var": "min_size",
```
mgfritch, line 201:
```
  >       assert args["var"] == "size"
  E       AssertionError: assert 'min_size' == 'size'
  E         - size
  E         + min_size
```

adjust the test case?

src/gravel/controllers/nodes/mgr.py
```diff
@@ -326,6 +333,20 @@ def gen() -> str:
 
         self._load()
 
+    async def _finish_bootstrap_config(self) -> None:
+        mon: Mon = Mon()
+        try:
+            mon.set_allow_pool_size_one()
+        except CephCommandError as e:
+            logger.error("=> mgr -- unable to allow pool size 1")
+            logger.debug(str(e))
```
jhesketh, line 342: (If you use `logger.exception` the logger will print the traceback along with the error string etc instead of needing to manually log it (but I realise you're using the debug level so this might not be helpful here)).

jhesketh (commented):
lgtm except for the test case

src/gravel/controllers/orch/ceph.py
jecluis: forgot about that. will do.

src/gravel/controllers/nodes/mgr.py
jecluis: huh - had no idea. That's really helpful. Although in this case I think we might just want the reason and not the whole stack trace.

jhesketh (approved):
Thanks for adding the typing to the fixtures :-)

PR #295, opened 2021-03-15
gravel: log to console and file; piggyback and relicense stuff

jecluis:
With this patchset we are

1. introducing automatic logging to file. The log is currently being placed under `/tmp/aquarium.log`, and it's limited to 10MB.
2. providing an endpoint to obtain the log file.
3. changing a bunch of logging messages to make output not look horrible.
4. relicensing a bunch of files to gplv3

### DIFF
--- a/src/gravel/controllers/nodes/conn.py
+++ b/src/gravel/controllers/nodes/conn.py
@@ -61,13 +61,13 @@ class ConnMgr:
         conn: IncomingConnection,
         msg: MessageModel
     ) -> None:
-        logger.debug(f"=> connmgr -- incoming recv: {conn}, {msg}")
+        logger.debug(f"connmgr -- incoming recv: {conn}, {msg}")
 
         if not self.is_started():
             raise ConnectionManagerNotStarted()
 
         await self._incoming_queue.put((conn, msg))
-        logger.debug(f"=> connmgr -- queue len: {self._incoming_queue.qsize()}")
+        logger.debug(f"connmgr -- queue len: {self._incoming_queue.qsize()}")
 
     async def wait_incoming_msg(
         self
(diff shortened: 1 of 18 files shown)


src/gravel/controllers/nodes/conn.py
```diff
@@ -61,13 +61,13 @@ async def on_incoming_receive(
         conn: IncomingConnection,
         msg: MessageModel
     ) -> None:
-        logger.debug(f"=> connmgr -- incoming recv: {conn}, {msg}")
```
kshtsk, line 64: instead of adding the decoration to the log, why just not add a module and function to the debug log format?

src/gravel/controllers/nodes/mgr.py
```diff
@@ -139,7 +139,7 @@ def __init__(self):
         self._node_init()
         assert self._state
 
-        logger.debug(f"=> mgr -- init > {self._state}")
+        logger.debug(f"init > {self._state}")
```
kshtsk, line 142: when logging data, it would be great to reader see what is it, in this case some info that this is a state be enough I guess.

src/gravel/controllers/nodes/conn.py
kshtsk: line number whould good as well

jecluis: we get a lot of pollution in the log. I tried that, and the logs become quite unreadable.

src/gravel/controllers/nodes/mgr.py
jecluis: Believe me, when it shows in the log you know it's state ;)

asettle (dismissed):

votdev (approved):
LGTM

kshtsk:
jenkins test tumbleweed

src/gravel/controllers/nodes/mgr.py
kshtsk: I am fine with that.

src/gravel/controllers/nodes/conn.py
kshtsk: Thanks, I see, makes sense. I know debug logs have a lot of garbage, this is why applications in normal mode do not print debug messages.
(This comment was not blocking the PR)

src/gravel/controllers/nodes/mgr.py
kshtsk: It is non-blocking comment, feel free to merge when ready.

asettle (approved):

PR #311, opened 2021-03-18
glass: Add services page

votdev:
Bildschirmfoto vom 2021-03-18 16-48-25

This is the core service page. In follow up PRs the ability to edit/delete them will be added. It is also planned to enhance the visualized data per service.

### DIFF
--- a/src/glass/src/app/app-routing.module.ts
+++ b/src/glass/src/app/app-routing.module.ts
@@ -10,6 +10,7 @@ import { DeploymentPageComponent } from '~/app/pages/deployment-page/deployment-
 import { InstallModePageComponent } from '~/app/pages/install-mode-page/install-mode-page.component';
 import { NotFoundPageComponent } from '~/app/pages/not-found-page/not-found-page.component';
 import { RegisterPageComponent } from '~/app/pages/register-page/register-page.component';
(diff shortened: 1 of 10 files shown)


asettle:
This looks really good!

jecluis (dismissed):
LGTM - although I do have concerns about "Usable size" instead of "Allocated size", but I think this is a broader discussion we need to have sooner rather than later.

votdev:
> LGTM - although I do have concerns about "Usable size" instead of "Allocated size", but I think this is a broader discussion we need to have sooner rather than later.

I'm fine with changing that.

s0nea (approved):
LGTM!

I'm only wondering if _CephFS Share_ should be renamed into _CephFS Service_ (just to be consistent with the wording).

Screenshot_2021-03-29_13-29-25

Also I created https://github.com/aquarist-labs/aquarium/issues/345 which can be addressed separately in my opinion.

votdev:
> LGTM!
> 
> I'm only wondering if _CephFS Share_ should be renamed into _CephFS Service_ (just to be consistent with the wording).
> 
> Screenshot_2021-03-29_13-29-25
> 
> Also I created #345 which can be addressed separately in my opinion.

I agree we have to polish the wording here and there in the UI. I took over the ' CephFS Share` from the services dashboard widget. So we have to adapt it there,too.

PR #312, opened 2021-03-18
.github/workflows: add glass unit tests config

kshtsk:
Use `npm run test:ci`

### DIFF
--- a/.github/workflows/glass-unit-tests.yml
+++ b/.github/workflows/glass-unit-tests.yml
@@ -1,0 +1,24 @@
+name: "Glass unit tests"
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  nodejs:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [14.x, 15.x]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - run: npm install
+        working-directory: src/glass
+      - run: npx ng build
+        working-directory: src/glass
+      - run: npm run test:ci
+        working-directory: src/glass


kshtsk:
This add node 14 and node 15 "glass unit test" action with 'test-ci' job.

kshtsk:
@jhesketh maybe there is more sense instead of this:

```
glass unit tests / test-ci
tox / build
```
try and use this:
```
ci / tox
ci / glass
```

jecluis (commented):
looks good to me, but I'll defer to @jhesketh

.github/workflows/glass-unit-tests.yml
```diff
@@ -0,0 +1,24 @@
+name: ci
```
jhesketh, line 1: Lets give this a more descriptive name so that on the "Checks" tab it's clear what this is.

Maybe "NPM Tests"?

jhesketh (changes requested):
I'm not the most familiar with npm or gh check's, but this looks good to me and the output looks good also. Just one small nit.

.github/workflows/glass-unit-tests.yml
kshtsk: @jhesketh npm is the package manager for node.js, the actual test command under the hood is _jest_ which is npm package. In other words npm is used not only for running tests but for building and packaging. Why ci / glass is not descriptive enough?

kshtsk: @jhesketh maybe we can merge it as is, and in upcoming PR you can play with the name you like more as for Checks?

jecluis: I'm not particularly invested in this discussion, but, maybe, then naming it "glass unit tests" would suffice?

jhesketh: Right, my example given is clearly poor, but my point was to have something other than CI because we have other systems also doing tests so we should differentiate them. "Glass unit tests" makes sense to me.

kshtsk:
jenkins run tumbleweed

jhesketh (approved):
Thanks for the update :-)

PR #313, opened 2021-03-18
gravel: fix bugs on bootstrap and join

### LINKED ISSUE #309: gravel: initial node requires aquarium restart after bootstrap before join is permitted
I'm not entirely sure how to describe the bug, aside from what happens: after we bootstrap the initial node, join attempts from another node result in a closed websocket with an invalid status code. Restarting Aquarium on the initial node makes everything okay without an itch.

At this point I have no concrete ideas about what might be wrong, but I suspect it has something to do with the node mgr / connection mgr rejecting a connection depending on state. Some state must be wrong somewhere, or we're simply not starting something.

jecluis:
Resolves #309 

### DIFF
--- a/src/gravel/controllers/bootstrap.py
+++ b/src/gravel/controllers/bootstrap.py
@@ -110,7 +110,7 @@ class Bootstrap:
         self.stage = BootstrapStage.RUNNING
 
         def progress_cb(percent: int) -> None:
-            logger.debug(f"bootstrap > {percent}")
+            logger.debug(f"bootstrap > {percent}%")
             self.progress = percent
 
         retcode: int = 0
--- a/src/gravel/controllers/nodes/mgr.py
+++ b/src/gravel/controllers/nodes/mgr.py
@@ -14,13 +14,11 @@
 import asyncio
 from enum import Enum
 from logging import Logger
-from datetime import datetime as dt
 from uuid import UUID, uuid4
 import random
 from typing import (
     Dict,
     Optional,
-    List
 )
 from pathlib import Path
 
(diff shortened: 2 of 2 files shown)


src/gravel/controllers/bootstrap.py
```diff
@@ -110,7 +110,7 @@ async def _do_bootstrap(self) -> None:
         self.stage = BootstrapStage.RUNNING
 
         def progress_cb(percent: int) -> None:
-            logger.debug(f"bootstrap > {percent}")
+            logger.debug(f"bootstrap > {percent}%")
```
mgfritch, line 113:
```suggestion
            logger.debug(f"bootstrap {percent}%")
```
(nit)

src/gravel/controllers/nodes/mgr.py
```diff
@@ -467,6 +431,31 @@ def _load_token(self) -> Optional[str]:
...
 
+    def _save_token(self, should_exist: bool = True) -> None:
+        tokenfile: Path = self._get_node_file("token")
+
+        # this check could be a single assert, but it's important to know which
+        # path failed.
+        if should_exist:
+            assert tokenfile.exists()
+        else:
+            assert not tokenfile.exists()
```
mgfritch, line 442: I'm confused, do we really need to assert this?

mgfritch (approved):
couple nits, but lgtm!

src/gravel/controllers/nodes/mgr.py
jecluis: We don't need to assert it, but it sure comes handy to ensure we are getting the right state at the right time from the caller's PoV. The alternative would be having the caller perform the asserts first, but that would require them to know where the file sits and hence increase boilerplate quite a bit.

PR #318, opened 2021-03-19
tools: setup-dev.sh: ensure we exit 0 on '--help'

jecluis:


ml8mr (approved):
Works for me.

PR #319, opened 2021-03-21
tests/vagrant: listen all host ips for older vagrant versions

kshtsk:
According to the recent docs of vagrant:

  host_ip (string) - The IP on the host you want to bind the forwarded port to.
  If not specified, it will be bound to every IP. By default, this is empty.

This is not the case for Leap 15.2 where vagrant version is 2.2.7 and
forwarded ports are bound only to localhost ip addresses.
While Tumbleweed has vagrant version 2.2.14 and does not have this
issue.


tserong (approved):
I assume this is actually a bug in vagrant 2.2.7, but sure, LGTM

PR #321, opened 2021-03-23
tool: add tool to deal with vagrant deployments, drop 'setup.sh'

jecluis:
Inspired by sesdev.

Hopefully will consolidate things a bit more for us, and especially newcomers, than a bunch of scripts spread throughout.

Stretch goal, eventually, will be to also get image builds going through this script.


l-mb (approved):
I like how this consolidates the code and makes the test environment more streamlined. Thanks for improving the docs as well!

(I've not actually tested it, but the docs/code look good to me.)

.wasser/config.yaml
```diff
@@ -55,10 +55,9 @@ routines:
         sudo systemctl start libvirtd
     - name: run microos image on vagrant
       command: |
-        cd aquarium/tests/vagrant
-        ./setup.sh foo 2>&1
-        cd setups/foo
-        vagrant up --no-destroy-on-error --no-parallel 2>&1
+        cd aquarium
+        ./tools/aqrdev create foo
```
kshtsk, line 59: add `2>&1` at the end to see why it is failing...

jecluis:
@kshtsk mind helping me out here? This is failing because of a `cd aquarium`, it seems, and I don't exactly know why. I tried mimicking the previous behavior, but maybe it was wrong?

.wasser/config.yaml
jecluis: ack

.wasser/config.yaml
```diff
@@ -55,10 +55,9 @@ routines:
...
+        ./tools/aqrdev start foo --conservative
```
kshtsk, line 60: and here 2>&1

tests/vagrant/templates/Vagrantfile.tmpl
```diff
@@ -2,15 +2,15 @@
...
     # The nfs server on Leap has udp enabled by default, but not on Tumbleweed
     # Disable udp to make it working for any host system by default.
-    config.vm.synced_folder "{{ROOTDIR}}", "/srv/aquarium", type: "nfs", nfs_udp: false
+    config.vm.synced_folder "{ROOTDIR}", "/srv/aquarium", type: "nfs", nfs_udp: false
     config.vm.synced_folder ".", "/vagrant", disabled: true
 
     config.vm.define :"node01", primary: true do |node|
         node.vm.hostname = "node01"
-        node.vm.network "forwarded_port", guest: 1337, host: 1337, host_ip: "*"
+        node.vm.network "forwarded_port", guest: 1337, host: 1337
```
kshtsk, line 13: why do you remove this?

tests/vagrant/templates/Vagrantfile.tmpl
```diff
@@ -24,7 +24,7 @@ Vagrant.configure("2") do |config|
 
     config.vm.define :"node02" do |node|
         node.vm.hostname = "node02"
-        node.vm.network "forwarded_port", guest: 1337, host: 1338, host_ip: "*"
+        node.vm.network "forwarded_port", guest: 1337, host: 1338
```
kshtsk, line 27: and this?

jecluis: isn't that something you changed recently? I may still need to rebase this on top of main.

tests/vagrant/templates/Vagrantfile.tmpl
jecluis: I didn't. must have been the move or something. I'll check it out.

tests/vagrant/templates/Vagrantfile.tmpl
kshtsk: yes... so you need to rebase then

kshtsk:
```
+ cd aquarium
+ ./tools/aqrdev create foo 2>&1
+ ./tools/aqrdev start foo --conservative 2>&1
>>> bash: ./tools/aqrdev: python3: bad interpreter: No such file or directory
>>> bash: ./tools/aqrdev: python3: bad interpreter: No such file or directory
ERROR:Failed to create node
```

tools/aqrdev
```diff
@@ -0,0 +1,805 @@
+#!python3
```
kshtsk, line 1: this odd

jecluis: wth

kshtsk:
that's better
but 'click' dependency is not resolved

jecluis:
> that's better
> but 'click' dependency is not resolved

this should have been met by uvicorn, which depends on click, but alright, I'll add that to the requirements.txt :)

jecluis:
> > that's better
> > but 'click' dependency is not resolved
> 
> this should have been met by uvicorn, which depends on click, but alright, I'll add that to the requirements.txt :)

nope, it's already there.

.wasser/config.yaml
```diff
@@ -55,10 +55,9 @@ routines:
         sudo systemctl start libvirtd
     - name: run microos image on vagrant
       command: |
-        cd aquarium/tests/vagrant
-        ./setup.sh foo 2>&1
-        cd setups/foo
-        vagrant up --no-destroy-on-error --no-parallel 2>&1
+        cd aquarium
+        ./tools/aqrdev create foo 2>&1
```
kshtsk, line 59: Suggestion why not make 'aqrdev' to work without a need to change dir to  aquarium?

jecluis:
Oh, I see. The tests are not running out of a virtualenv. I'm unsure if we want to have this stuff being installed in the user's machine just to do development work -- after all, that's what the virtualenv is there for. Any thoughts?

.wasser/config.yaml
jecluis: at this point it requires to have the git repository somewhere as a parent, otherwise it will break. That's something I'm not entirely sure how to fix without having the user specifying the root of the repository.

kshtsk:
What about use name `acrylic` instead of `aqrdev` and move it from tools to a separate directory?

jecluis:
> What about use name `acrylic` instead of `aqrdev` and move it from tools to a separate directory?

I'm not opposed to renaming it and putting it in a different directory, but I'm not seeing how that will fix the dependency/virtualenv issue?

kshtsk:
> > What about use name `acrylic` instead of `aqrdev` and move it from tools to a separate directory?
> 
> I'm not opposed to renaming it and putting it in a different directory, but I'm not seeing how that will fix the dependency/virtualenv issue?

it is not related suggestion to the dependency issue

jecluis:
> > > What about use name `acrylic` instead of `aqrdev` and move it from tools to a separate directory?
> > 
> > 
> > I'm not opposed to renaming it and putting it in a different directory, but I'm not seeing how that will fix the dependency/virtualenv issue?
> 
> it is not related suggestion to the dependency issue

Alright, Sorry ;)

.wasser/config.yaml
kshtsk: I usually calculate the root of repo from the __file__ path as a relation.

kshtsk:
> Oh, I see. The tests are not running out of a virtualenv. I'm unsure if we want to have this stuff being installed in the user's machine just to do development work -- after all, that's what the virtualenv is there for. Any thoughts?

drop using `click` and standard `argparse`?

jecluis:
> > Oh, I see. The tests are not running out of a virtualenv. I'm unsure if we want to have this stuff being installed in the user's machine just to do development work -- after all, that's what the virtualenv is there for. Any thoughts?
> 
> drop using `click` and use the standard `argparse`?

This does not solve the wider problem of not being able to add new dependencies to this script or any other dev script the CI infra might use.

jecluis:
> but if you're asking me if we need another dependency resolution for the new tool, then I don't know what you had in mind dropping the setup-dev.sh script.

I didn't drop the `setup-dev.sh` script. I dropped the `setup.sh` script and replaced it with this.

> So we either need to add to FZTH doc how to use this tool. Btw, when you move your python tool to a separate directory there can be a setup config where you can provide all dependencies for it and you just need to install it to virtual environment as a develop target, the same as I did for wasser, but it is probably not what you want

What about, instead of doing this kind of foo, we activate the venv to run this command in the wasser yaml, and deactivate afterwards? We should have a venv as setup earlier by the `setup-dev.sh` script, right?

.wasser/config.yaml
```diff
@@ -55,10 +55,11 @@ routines:
...
       command: |
-        cd aquarium/tests/vagrant
-        ./setup.sh foo 2>&1
-        cd setups/foo
-        vagrant up --no-destroy-on-error --no-parallel 2>&1
+        cd aquarium
+        source venv/bin/activate
+        ./tools/aqrdev create foo 2>&1
+        ./tools/aqrdev start foo --conservative 2>&1
+        deactivate
```
kshtsk, line 62: You do not need to deactivate it since, this channel is dropped, every new command needs own env

kshtsk:
> What about, instead of doing this kind of foo, we activate the venv to run this command in the wasser yaml, and deactivate afterwards? We should have a venv as setup earlier by the `setup-dev.sh` script, right?

that makes sense to me

.wasser/config.yaml
jecluis: yeah, that is one of the options, but I'm not so sure how that would play out if the file is moved in the future. Something to kick down the road and think about later on.

.wasser/config.yaml
jecluis: alright, will adjust.

kshtsk: yeah... for illustration it looks okay to leave it here to correspond steps in FZTH

.wasser/config.yaml
kshtsk: well, when you move the file around, you tweak the relational path add or remove `..` :-D

.wasser/config.yaml
jecluis: Does this mean this command does not have a virtualenv without running `./setup-dev.sh` first?

kshtsk:
but still, we need to add click to the dev venv

kshtsk:
or just python3-click to system?

jecluis:
> but still, we need to add click to the dev venv

it's already there: https://github.com/aquarist-labs/aquarium/blob/main/src/requirements.txt#L2

kshtsk:
Ah... I see, because you put "#!/bin/python3" or something it always takes system python3 instead of venv. I suggest you to replace it with `#!env python3`

kshtsk:
sorry, I meant `#!/usr/bin/env python3`

kshtsk:
now it is not clear how to ssh to a box and run commands in it?

jecluis:
> now it is not clear how to ssh to a box and run commands in it?

no, it was a mistake. the location of the deployments changed and I forgot to change that in the wasser config.

jecluis:
Hurray!

kshtsk:
excellent

tserong (approved):
As Lars said, I like that this simplifies running Vagrant. With the caveat that I haven't actually run it myself, but that the tests all pass, LGTM :-)

PR #326, opened 2021-03-23
gravel: simplify unknown service type handling

mgfritch:
handle unknown service types in a generic way, rather than a
cephfs specific way

### DIFF
--- a/src/gravel/controllers/services.py
+++ b/src/gravel/controllers/services.py
@@ -35,10 +35,10 @@ class NotEnoughSpaceError(ServiceError):
 
 class ServiceTypeEnum(str, Enum):
     CEPHFS = "cephfs"
-    NFS = "nfs"
-    RBD = "rbd"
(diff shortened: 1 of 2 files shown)


src/gravel/controllers/services.py
```diff
@@ -146,8 +146,7 @@ def check_requirements(
     def _create_service(self, svc: ServiceModel) -> None:
         if svc.type == ServiceTypeEnum.CEPHFS:
             self._create_cephfs(svc)
-        else:
-            raise NotImplementedError("only cephfs is currently supported")
+        raise NotImplementedError(f"unknown service type: {svc.type}")
```
ml8mr, line 149: Shouldn't this statement be under the else block?

mgfritch: oops! yes ... fixed :)

ml8mr (approved):
LGTM

jecluis (approved):

Issue #277, opened 2021-03-10
Add a Makefile


tserong:
Just thinking about our build scripts while looking at #91, it might be nice to refactor things slightly so we have a Makefile, and can e.g.: do `make image` as opposed to `./tools/build-image.sh`, add a `make dist` for getting a tarball for package building, etc.

Issue #91, opened 2021-02-10
aquarium: package glass and gravel


jecluis:
**Context**
Distributing our mighty neat appliance in a ready to consume way is a good path to adoption. Containerizing it will allow users to go through a very simple process to get Aquarium up and running.

**Describe the solution you'd like**
A container containing glass and gravel, with gravel running upon start and serving glass.

**Additional context**
Something that I personally still struggle with is how the bootstrapping will happen. Given we're running on a container, how are we going to start other containers on the host?

tserong:
> Something that I personally still struggle with is how the bootstrapping will happen. Given we're running on a container, how are we going to start other containers on the host?

That would be the trick... Containers aren't really meant to be able to take over the host and tell it to do stuff, right?

I wonder if you could do something ~evil~ ~abominable~ clever with bind mounts inside the container?

Or, as another alternative, we _don't_ put aquarium in a container; instead we build a regular package (RPM, DEB, etc.) then in out MicroOS case, we just install the aquarium RPM (and it can then be updated the same as the rest of MicroOS, as a package). If someone wants to deploy aquarium on, say, Debian, they install our DEB. Of course the downside there is we end up maintaining a handful of different packages for different distros vs. a single container image. But it _would_ fix the bootstrap problem ;-)

tserong:
...although, it's really only calls to `cephadm` we care about, right? Everything else will go through `ceph orch` or librados or whatever. We could make the aquarium container ssh to the host to invoke `cephadm boostrap` (given all the hosts need ssh running anyway, because that's how cephadm itself works).

asettle:
@theanalyst any update on this?

sebastian-philipp:
> > Something that I personally still struggle with is how the bootstrapping will happen. Given we're running on a container, how are we going to start other containers on the host?
> 
> That would be the trick... Containers aren't really meant to be able to take over the host and tell it to do stuff, right?
> 
> I wonder if you could do something ~evil~ ~abominable~ clever with bind mounts inside the container?

calling executables on the host (like cephadm) from within a container is a bit awkward. That's why we opted for a non-containerized cephadm daemon.  

> 
> Or, as another alternative, we _don't_ put aquarium in a container; instead we build a regular package (RPM, DEB, etc.) then in our MicroOS case, we just install the aquarium RPM (and it can then be updated the same as the rest of MicroOS, as a package). 

:+1: 

[...]

PR #332, opened 2021-03-25
Introduce runtime python venv for dependencies, add Makefile

tserong:
This is work towards being able to have an actual package (rpm, deb, ...) of Aquarium (see https://github.com/aquarist-labs/aquarium/issues/91).

I've added a Makefile with a `dist` target, so you can now run `make dist` to get a tarball named something like aquarium-0.1.0-0-gc28eb76.tar.gz, which includes a built glass and all the necessary bits of gravel.  It also includes all our python dependencies (per src/requirements.txt) in /usr/share/aquarium/venv, so we don't have to worry about not having packaged versions of these things on whatever distros we deploy on.

This doesn't change the current dev workflow at all.  You still build an image by running `./tools/build-image.sh`, but that script will in turn call `make dist` to get a tarball to install in the image.  In future (once I've got a spec file and rpm build sorted out), this can be changed so that the kiwi config uses that actual package, instead of the tarball.

That said, if you *don't* want to build an image, but would rather try a manual Aquarium install on some random Linux, you could run `make dist`, then take the tarball and extract it on the system you want to deploy on.  Might be a bit of fun ;-)

Note: the version string used in the tarball name is based on the latest annotated tag in the form 'vX.Y.Z', which means I've suddenly introduced the notion of actual version numbers and releases to the project.  Sorry.

Fixes: https://github.com/aquarist-labs/aquarium/issues/277

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,59 @@
+# The version needs an annotated tag in the git repo in the form vX.Y.Z.
+# If such is not present, fall back to using v0.0.0 so we get a non-empty
+# version string.
+VERSION=$(shell (git describe --match 'v*' --long 2>/dev/null || echo 'v0.0.0') | sed 's/^v//')
+
+usage:
+	@echo "Try running \`make dist\` to build a tarball"
+
+submodules:
+	git submodule update --init
+
+glass:
+	cd src/glass && npm install && npx ng build --prod --output-hashing=all
+
+dist: submodules glass
+	$(eval TMPDIR := $(shell mktemp -d))
+	$(eval TAR_BASE := $(TMPDIR)/aquarium-$(VERSION))
+	$(eval TAR_USR := $(TAR_BASE)/usr/share/aquarium)
+	$(eval TAR_UNIT := $(TAR_BASE)/usr/lib/systemd/system)
+	$(eval TAR_SBIN := $(TAR_BASE)/usr/sbin)
+	mkdir -p $(TAR_USR) $(TAR_UNIT) $(TAR_SBIN)
+	# Copy gravel, glass, aquarium.py and cephadm from src/...
+	cd src && \
+	find gravel -iname '*ceph.git*' -prune -false -o -iname '*.py' | \
+		xargs cp --parents --target-directory=$(TAR_USR) && \
+	cp --target-directory=$(TAR_USR) ./aquarium.py && \
+	cp -R --parents --target-directory=$(TAR_USR) glass/dist && \
+	cp --target-directory=$(TAR_SBIN) ./gravel/ceph.git/src/cephadm/cephadm
+	# Copy aquarium service
+	cp systemd/aquarium.service $(TAR_UNIT)
+	# Create python venv for use by aquarium at runtime.  Have to explicitly
+	# specify /usr/bin/python3 here, rather than just calling python3 with
+	# no path, otherwise if we're already in a venv when running `make dist`,
+	# /usr/share/aquarium/venv/bin/python3 will point to the existing venv
+	# python on the build host, which is completely wrong.
+	/usr/bin/python3 -m venv $(TAR_USR)/venv
+	# Install requirements (note: deliberately *not* using
+	# --system-site-packages, otherwise we might accidentally not install
+	# some of our dependencies which are already present on the build host)
+	$(TAR_USR)/venv/bin/pip install -r src/requirements.txt
+	# At this point, the shebang lines for everyting in venv/bin will be
+	# absolute paths on the build host (e.g. /tmp/tmp.QoQm2t2j06/aquarium-0.1.0-0-gc28eb76/usr/share/aquarium/venv/bin/python3)
+	# but we need them to be appropriate for the host we're *installing* on,
+	# so need to strip out the leading part of the path.  Also, even though
+	# the venv was created without --system-site-packages, we actually need
+	# to set that to true now, so that python *inside* the venv will find
+	# the python rados library which is installed as a package.
+	cd $(TAR_USR)/venv ; \
+	fixfiles=$$(find -type f -not -name *.py[cox] -exec grep -Il "#\!.*$(TAR_BASE)" {} \;) ; \
+	fixfiles="$$fixfiles bin/activate*" ; \
+	for f in $$fixfiles; do \
+		echo -n "fixing path in $$f: " ; \
+		grep $(TAR_BASE) "$$f" ; \
+		sed -i -e "s;$(TAR_BASE);;" $$f ; \
+	done ; \
+	sed -i -e "s;^include-system-site-packages.*;include-system-site-packages = true;" pyvenv.cfg
+	tar -czf aquarium-$(VERSION).tar.gz -C $(TMPDIR) aquarium-$(VERSION)
+	rm -r $(TMPDIR)
+
--- a/images/microOS/config.xml
+++ b/images/microOS/config.xml
@@ -603,12 +603,8 @@
         <package name="lvm2"/>
         <package name="chrony"/>
         <package name="tuned"/>
-        <package name="python3-pip"/>
-        <package name="python3-aiofiles"/>
-        <package name="python3-requests"/>
         <package name="python3-rados"/>
-        <package name="python3-websockets"/>
-        <archive name="aquarium.tar"/>
+        <archive name="aquarium.tar.gz"/>
     </packages>
 
     <packages type="image" profiles="kvm-and-xen,kvm-and-xen_x86_64,kvm-and-xen_aarch64,VMware,MS-HyperV,VirtualBox,Pine64,RaspberryPi,RaspberryPi2,Vagrant_x86_64,Vagrant_aarch64">
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,4 @@ tests/vagrant/deployments/
 .vagrant/
 .idea/
 conf/
-
+*.tar.gz
--- a/images/microOS/config.sh
+++ b/images/microOS/config.sh
@@ -174,7 +174,6 @@ EOF
 fi
 
 if [[ "$kiwi_profiles" == *"Ceph"* ]]; then
-  pip install fastapi uvicorn
   baseInsertService aquarium
 fi
 
--- a/systemd/aquarium.service
+++ b/systemd/aquarium.service
@@ -5,7 +5,7 @@ After=network.target
 [Service]
 User=root
 WorkingDirectory=/usr/share/aquarium
-ExecStart=/usr/local/bin/uvicorn aquarium:app --host 0.0.0.0 --port 1337
+ExecStart=/usr/share/aquarium/venv/bin/uvicorn aquarium:app --host 0.0.0.0 --port 1337
 Restart=always
 
 [Install]
--- a/tools/build-image.sh
+++ b/tools/build-image.sh
@@ -131,63 +131,24 @@ fi
 
 set -xe
 
-glassdir=${srcdir}/glass
-graveldir=${srcdir}/gravel
-
-[[ ! -d "${glassdir}" ]] && \
-  error_exit "missing frontend directory at '${glassdir}"
-
-[[ ! -e "${glassdir}/angular.json" ]] && \
-  error_exit "directory at '${glassdir}' is not an angular application"
-
-[[ ! -d "${graveldir}" ]] && \
-  error_exit "missing backend directory at '${graveldir}"
-
-
-build_glass() {
-  pushd ${glassdir}
-  npm install || exit 1
-  npx ng build --prod --output-hashing=all || exit 1
-  [[ ! -d "${glassdir}/dist" ]] && \
-    error_exit "missing 'dist' directory for frontend"
-  popd
-}
-
-bundle() {
-  local bundledir=${build}/bundle
-  local bundle_usr=${bundledir}/usr/share/aquarium
-  local bundle_unit=${bundledir}/usr/lib/systemd/system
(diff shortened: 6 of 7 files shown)


tserong:
I've marked this WIP because I don't know if the CI is going to like this or not, because I've introduced a requirement on having a version tag.

tserong:
Ah, dammit, I accidentally included src/glass/package-lock.json as well. That's a mistake. Will fix.

tserong:
Yep, this won't do the right thing in CI without a tag. Hang on...

jecluis:
I'm perfectly fine with adding a version tag to the repo, at the current HEAD, for whatever. This is the first time we've needed a version, so might as well make it a major milestone and give in ;)

tserong:
> (to be clear, it should be a minor version, like 0.0.1 or whatever)

Yeah, totally :-) https://semver.org/ suggests v0.1.0 for that case. But this PR is necessarily a bit chicken-and-eggy, as it (ideally) wants a release tag, except there isn't one yet. So I've made it fall back to a hardcoded v0.0.0 if `git describe` fails for whatever reason.

tserong:
OK, almost there, just need to figure out why setup-vagrant-tumbleweed succeeded but setup-vagrant-leap failed...

jecluis:
> > (to be clear, it should be a minor version, like 0.0.1 or whatever)
> 
> Yeah, totally :-) https://semver.org/ suggests v0.1.0 for that case. But this PR is necessarily a bit chicken-and-eggy, as it (ideally) wants a release tag, except there isn't one yet. So I've made it fall back to a hardcoded v0.0.0 if `git describe` fails for whatever reason.

Yeah, 0.1.0 works, whatever ;) Meant as an example.

Makefile
```diff
@@ -0,0 +1,59 @@
...
+	# /usr/share/aquarium/venv/bin/python3 will point to the existing venv
+	# python on the build host, which is completely wrong.
+	/usr/bin/python3 -m venv $(TAR_USR)/venv
+	# Install requirements (note: deliberately *not* using
+	# --system-site-packages, otherwise we might accidentally not install
+	# some of our dependencies which are already present on the build host)
+	$(TAR_USR)/venv/bin/pip install -r src/requirements.txt
+	# At this point, the shebang lines for everyting in venv/bin will be
+	# absolute paths on the build host (e.g. /tmp/tmp.QoQm2t2j06/aquarium-0.1.0-0-gc28eb76/usr/share/aquarium/venv/bin/python3)
+	# but we need them to be appropriate for the host we're *installing* on,
```
jecluis, line 43: When building the image, this is something that can be achieved by the `disk.sh` script, which is run inside the chroot'ed environment just before the image is built. So, technically, one could package the `requirements.txt` file in the tar, drop it somewhere we could reach it within the `chroot`, create the `venv` and install the packages, and then remove the `requirements.txt` file.

Otoh, that would not work for the bare metal host :shrug:

Makefile
```diff
@@ -0,0 +1,59 @@
...
+	# so need to strip out the leading part of the path.  Also, even though
+	# the venv was created without --system-site-packages, we actually need
+	# to set that to true now, so that python *inside* the venv will find
+	# the python rados library which is installed as a package.
+	cd $(TAR_USR)/venv ; \
+	fixfiles=$$(find -type f -not -name *.py[cox] -exec grep -Il "#\!.*$(TAR_BASE)" {} \;) ; \
+	fixfiles="$$fixfiles bin/activate*" ; \
+	for f in $$fixfiles; do \
+		echo -n "fixing path in $$f: " ; \
+		grep $(TAR_BASE) "$$f" ; \
```
jecluis, line 53: meant to continue here? what am I missing?

jecluis (commented):
LGTM minus a worry on a `for` during path replacement.

Will test before giving the a-okay.

Makefile
tserong: The grep just results in printing out the path it found that it's changing...  Maybe it's not necessary?  This whole "fixfiles" bit of evil^H^H^H cleverness came from Calamari :-)

tserong:
From the logs (https://ceph-ci.suse.de/job/aquarium-setup-vagrant-leap/74/consoleFull), I don't think I can tell why it's failing; all I can see is:

```=== check aquarium service is up and running
+ cd aquarium/tests/vagrant/deployments/foo
+ vagrant ssh -c "sudo systemctl status aquarium"
>>> ��� aquarium.service - Project Aquarium
>>>      Loaded: loaded (/usr/lib/systemd/system/aquarium.service; enabled; vendor preset: disabled)
>>>      Active: failed (Result: exit-code) since Thu 2021-03-25 08:05:31 UTC; 1min 22s ago
>>>    Main PID: 1393 (code=exited, status=1/FAILURE)
>>> 
>>> Mar 25 08:05:31 localhost systemd[1]: aquarium.service: Scheduled restart job, restart counter is at 5.
>>> Mar 25 08:05:31 localhost systemd[1]: Stopped Project Aquarium.
>>> Mar 25 08:05:31 localhost systemd[1]: aquarium.service: Start request repeated too quickly.
>>> Mar 25 08:05:31 localhost systemd[1]: aquarium.service: Failed with result 'exit-code'.
>>> Mar 25 08:05:31 localhost systemd[1]: Failed to start Project Aquarium.
```

@kshtsk I guess I need to somehow add `journalctl -u aquarium` or similar if `systemctl status aquarium` fails at this point, so we can see the full log of what went wrong?

kshtsk:
@tserong you can probably add -n 100 to `systemctl status`?

images/microOS/config.xml
```diff
@@ -603,12 +603,8 @@
         <package name="lvm2"/>
         <package name="chrony"/>
         <package name="tuned"/>
-        <package name="python3-pip"/>
-        <package name="python3-aiofiles"/>
-        <package name="python3-requests"/>
         <package name="python3-rados"/>
-        <package name="python3-websockets"/>
```
kshtsk, line 610: Why those dependencies are dropped?

jecluis: because we're now relying on a virtualenv to run stuff from. These dependencies are being installed in the virtualenv.

kshtsk: This is a wrong approach, if the venv is prepared on the host. The host system is not necessary the same as the guest one.

jecluis: Can you be more specific?

kshtsk: but I guess it is done in a right way

kshtsk:
```Mar 25 12:38:02 localhost uvicorn[1352]: Traceback (most recent call last):
Mar 25 12:38:02 localhost uvicorn[1352]:   File "/usr/share/aquarium/venv/bin/uvicorn", line 5, in <module>
Mar 25 12:38:02 localhost uvicorn[1352]:     from uvicorn.main import main
Mar 25 12:38:02 localhost uvicorn[1352]: ModuleNotFoundError: No module named 'uvicorn'
```
this is what is seen in journalctl

images/microOS/config.xml
kshtsk: > Can you be more specific?

for example, host system is Leap, guest system is Tumbleweed, there two different system package dependencies as well as python3 version itself, if you bundle venv on host system and try run it on guest one, you'll fail.

kshtsk: As we have failed for the Leap case in this particular PR.

kshtsk: Same true for guys who are trying this out on Ubuntu or RH hosts.

kshtsk:
```vagrant@node01:~> /usr/share/aquarium/venv/bin/uvicorn aquarium:app --host 0.0.0.0 --port 1337
Traceback (most recent call last):
  File "/usr/share/aquarium/venv/bin/uvicorn", line 5, in <module>
    from uvicorn.main import main
ModuleNotFoundError: No module named 'uvicorn'
vagrant@node01:~> . /usr/share/aquarium/venv/bin/activate
(venv) vagrant@node01:~> pip list
Traceback (most recent call last):
  File "/usr/share/aquarium/venv/bin/pip", line 5, in <module>
    from pip._internal.cli.main import main
ModuleNotFoundError: No module named 'pip'
```

kshtsk:
```vagrant@node01:~> python3 --version
Python 3.8.8
vagrant@node01:~> ls /usr/share/aquarium/venv/lib64
python3.6
```

System and venv python version mismatch.

kshtsk:
> @kshtsk I guess I need to somehow add `journalctl -u aquarium` or similar if `systemctl status aquarium` fails at this point, so we can see the full log of what went wrong?

that makes sense to me, like:
```
sudo systemctl status aquarium || {
   RET=$?
   sudo journalctl -u aquarium
   exit $RET
}
```

images/microOS/config.xml
tserong: They're embedded in the tarball now inside a python venv

tserong:
> System and venv python version mismatch.

Ah, damn, I may need to rethink this approach. The venv in the tarball is created on the build host, using whatever version of python is installed there, so in the Leap case, python 3.6. But then the tarball is installed inside the Tumbleweed-based MicroOS which uses python 3.8...

kshtsk:
> > System and venv python version mismatch.
> 
> Ah, damn, I may need to rethink this approach. The venv in the tarball is created on the build host, using whatever version of python is installed there, so in the Leap case, python 3.6. But then the tarball is installed inside the Tumbleweed-based MicroOS which uses python 3.8...

Maybe use dockerized build? Or run your build script in the vagrant instance? Keep in mind though major version may match it also matching minor version and even system dependencies.

images/microOS/config.xml
kshtsk: Why do we need to embed dependencies into tarball? Why not to use pip to install?

tserong: I want to ultimately make a package of aquarium (rpm, deb, etc.), so i'm making a venv inside the tarball and using pip to install in that venv when the tarball is built. The implication of this, which I apparently hadn't realised until yesterday when the leap CI failed, is that the the host you build the tarball on, basically has to be the same OS the tarball is deployed on. That will (or should) eventually work for the case of providing an aquarium package for any distro, but right now breaks when using any non-tumbleweed distro to build the MicroOS image.

tserong:
Marking WIP again, because I obviously have some more work to do here.

images/microOS/config.xml
kshtsk: @tserong maybe use kiwi for that?

jecluis:
@tserong do you thing we could split off the bundling of Aquarium's stuff for `make dist` without the pip stuff for now, and then deal with pip later? TBH, I find that quite useful as is and the more I think about how that could help us get rid of the `build-image.sh` script, the more I want it.

kshtsk:
Guys, I still don't get why do you want include all the dependencies in the tar ball, maybe it better to allow more natural way for python to deploy, like using pip so it will install all dependencies from the pypi? Otherwise we need to build rpms for each system? Alternatively, the package can be probably built via docker for the guest system.

jecluis:
> Guys, I still don't get why do you want include all the dependencies in the tar ball, maybe it better to allow more natural way for python to deploy, like using pip so it will install all dependencies from the pypi? Otherwise we need to build rpms for each system? Alternatively, the package can be probably built via docker for the guest system.

Tim is still looking into alternatives and the best path forward. What I'm suggesting is having the makefile to do the bundling of aquarium (read, gravel, glass, cephadm, etc), without dependencies, before we figure the rest out.

tserong:
> @tserong do you thing we could split off the bundling of Aquarium's stuff for `make dist` without the pip stuff for now, and then deal with pip later? TBH, I find that quite useful as is and the more I think about how that could help us get rid of the `build-image.sh` script, the more I want it.

Alright, I've done that in https://github.com/aquarist-labs/aquarium/pull/382 (I didn't rip the pip stuff out of this one as I'd still rather keep it around for posterity just in case).

> Guys, I still don't get why do you want include all the dependencies in the tar ball, maybe it better to allow more natural way for python to deploy, like using pip so it will install all dependencies from the pypi?

My goal was to try to create a native distro package of aquarium (rpm, deb, etc.) that could be installed on openSUSE, Fedora, Debian, Ubuntu. That means that we'd need all our python dependencies packaged for all those distros too. Some of them are packaged, and some are not, and I have _zero_ desire to take on package maintenance for a bunch of python packages on multiple Linux distros :-) If we use `pip` at _build_ time, to embed those dependencies inside the tarball, that dependency problem goes away. Does that help explain why I was experimenting with this approach?

kshtsk:
> My goal was to try to create a native distro package of aquarium (rpm, deb, etc.) that could be installed on openSUSE, Fedora, Debian, Ubuntu. That means that ordinarily we'd need all our python dependencies packaged for all those distros too. Some of them are packaged, and some are not, and I have _zero_ desire to take on package maintenance for a bunch of python packages on multiple Linux distros :-) 

Yeah, understandable.

> If we use `pip` at _build_ time, to embed those dependencies inside the tarball, that dependency problem goes away. Does that help explain why I was experimenting with this approach?

As you can see it does not go away but makes it even worth. This is why the packaging thing at _build_ time should be run probably in guest container with the destination os instead of the bare host.

tserong:
> As you can see it does not go away but makes it even worth.

Indeed :-/

> This is why the packaging thing at _build_ time should be run probably in guest container with the destination os instead of the bare host.

Yeah, that sounds like it'd be more reliable than my first attempt.

asettle:
Any update on this PR @tserong ?

tserong:
> Any update on this PR @tserong ?

This has been on the back burner for >checks notes< _way_ too long, so I should probably just close it.

PR #333, opened 2021-03-25
gravel: keep track of events

jecluis:
This is mostly meant for feedback gathering -- I'm not particularly enthused with it.

It's essentially a very dumb queue of items, added by other parts of the backend, which are supposed to be events.

Although we are doing a very, _very_ basic form of collision detection, we're not being smarter than that. Additionally, we're not providing context, we're not tracking how often a given event has been raised, and we're definitely not ordering them by severity when listing them.

It also seems a bit heavy, and is making my VMs a bit too slow. Might be the list reversal, or it might be that too much is happening in 1 second bursts -- and these are pretty light VMs.

Does anyone has thoughts on what this should look like?

### DIFF
--- a/src/aquarium.py
+++ b/src/aquarium.py
@@ -31,7 +31,8 @@ from gravel.api import (
     services,
     nodes,
     local,
-    devices
+    devices,
+    events
 )
 
 
@@ -66,6 +67,10 @@ api_tags_metadata = [
     {
         "name": "devices",
         "description": "Obtain and perform operations on cluster devices"
+    },
+    {
+        "name": "events",
+        "description": "Obtain list of latest events"
     }
 ]
 
@@ -107,6 +112,7 @@ api.include_router(status.router)
 api.include_router(services.router)
 api.include_router(nodes.router)
 api.include_router(devices.router)
+api.include_router(events.router)
 
 
 #
--- a/src/gravel/api/events.py
+++ b/src/gravel/api/events.py
@@ -1,0 +1,42 @@
+# project aquarium's backend
+# Copyright (C) 2021 SUSE, LLC.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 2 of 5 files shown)


tserong:
I assume the intent here is to expose "most recent instances of some sort of doom" ;-)

Could this alternately be done by scraping the cluster log somehow, rather that keeping our own event list? Or are we likely to want to know about other cluster-wide events that Ceph itself is unaware of?

Does the list need to be copied and reversed? If that's a bottleneck, could the caller be required to iterate through that list backwards instead?

votdev:
> I assume the intent here is to expose "most recent instances of some sort of doom" ;-)
> 
> Could this alternately be done by scraping the cluster log somehow, rather that keeping our own event list? Or are we likely to want to know about other cluster-wide events that Ceph itself is unaware of?
> 
> Does the list need to be copied and reversed? If that's a bottleneck, could the caller be required to iterate through that list backwards instead?

The UI wants to have the list in a reverse order because the user surely wants to see the latest event at the top of the list. But reverting the list can be done in the frontend. What is much more important is pagination.

s0nea:
Somehow I can't link the PR to the issue: https://github.com/aquarist-labs/aquarium/issues/387

asettle:
> This is mostly meant for feedback gathering -- I'm not particularly enthused with it.
> 
> It's essentially a very dumb queue of items, added by other parts of the backend, which are supposed to be events.
> 
> Although we are doing a very, _very_ basic form of collision detection, we're not being smarter than that. Additionally, we're not providing context, we're not tracking how often a given event has been raised, and we're definitely not ordering them by severity when listing them.
> 
> It also seems a bit heavy, and is making my VMs a bit too slow. Might be the list reversal, or it might be that too much is happening in 1 second bursts -- and these are pretty light VMs.
> 
> Does anyone has thoughts on what this should look like?
> 
> Signed-off-by: Joao Eduardo Luis <[joao@suse.com](mailto:joao@suse.com)>

Not to be that person about 3 months late, but perhaps this could be moved to the Discussion board or closed?

jecluis:
Oh yeah, definitely. This was simply, _ahem_, forgotten...

Issue #103, opened 2021-02-11
gravel: set ceph config options that allows us to have proper HEALTH_OK on a small cluster


jecluis:
**Is your feature request related to a problem? Please describe.**
A small cluster will always be in either HEALTH_WARN or HEALTH_ERR with the default options. Pools will not be available due to `min_size`, and will be degraded due to `size`.

**Describe the solution you'd like**
Adjust defaults so pools are created with basic requirements, adjust them to whatever they need upon creation.

asettle:
@ifed01 any update on this? :) Do you have any questions? We are looking to complete this by Milestone 2 (31 March)

ifed01:
> @ifed01 any update on this? :) Do you have any questions? We are looking to complete this by Milestone 2 (31 March)

not yet, will put in on the top of my TODO list

ifed01:
Well, the root cause for the undersized PGs is actually a wrong crushmap which requires per-host replication not per-osd one.

ifed01:
so we should either set 'osd_crush_chooseleaf_type' to 0 prior to cluster start up or edit crushmap(s) afterwards.

jecluis:
We will likely have to either allow ourselves to edit the crushmap anyway, or create multiple rules to be used depending on our current state. This because, while we start with just one node, we will eventually expand to a second node and the conditions then change.

I don't know whether we can rely on `osd_pool_default_crush_rule` to set the crush rule to something else besides the default `replicated_rule`; if so, we'd create a new rule for single-node and set that option to that new rule, and do that just after bootstrap but before allowing deploying the first osds.

And if that pans out, we can create a new ruleset when a second node is added, and then change pool's ruleset to the new ruleset.

Maybe?

ifed01:
> We will likely have to either allow ourselves to edit the crushmap anyway, or create multiple rules to be used depending on our current state. This because, while we start with just one node, we will eventually expand to a second node and the conditions then change.
> 
> I don't know whether we can rely on `osd_pool_default_crush_rule` to set the crush rule to something else besides the default `replicated_rule`; if so, we'd create a new rule for single-node and set that option to that new rule, and do that just after bootstrap but before allowing deploying the first osds.
> 
> And if that pans out, we can create a new ruleset when a second node is added, and then change pool's ruleset to the new ruleset.
> 
> Maybe?

Honestly this looks pretty complicated and rather unnecessary for the typical expected deployments which IMO gonna have 3+ nodes.

PR #341, opened 2021-03-28
gravel: make default pool size one

jecluis:
This is mostly meant to workaround #103, but opens all kinds of
questions. Without it though, nfs cluster creation will hang because the
'nfs-ganesha' pool will be unavailable on a single-node.


jecluis:
Superseeded by #342

PR #342, opened 2021-03-28
gravel: adjust crush rulesets depending on number of nodes

### LINKED ISSUE #103: gravel: set ceph config options that allows us to have proper HEALTH_OK on a small cluster

jecluis:
When we create a single-node deployment, we should enable the user to have a healthy system out of the box; with that in mind, we must ensure that our replicas are properly replicated across OSDs, instead of the default, hosts.

With this patchset we first ensure that we create a single-node ruleset, and set it as the default for new pools upon bootstrap and before we start creating pools; then we ensure that we reset the default to `replicated_rule` upon join of a new node, as well as adjusting the existing pools' rulesets.

Resolves #103 

### DIFF
--- a/src/gravel/controllers/orch/ceph.py
+++ b/src/gravel/controllers/orch/ceph.py
@@ -21,7 +21,8 @@ from typing import (
     Callable,
     Dict,
     Any,
-    List
+    List,
+    Optional
 )
 from logging import Logger
 from fastapi.logger import logger as fastapi_logger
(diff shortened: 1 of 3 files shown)


mgfritch (commented):
wondering about :
1. blindly discarding existing rulesets that could be user defined (or service dependent?)
2. what happens when we go from 2 nodes -> 1 node

maybe it's a non-issue since we are opinionated...

jecluis:
On Tue, Mar 30, 2021, 01:37 Michael Fritch ***@***.***> wrote:

> ***@***.**** commented on this pull request.
>
> wondering about :
>
>    1. blindly discarding existing rulesets that could be user defined (or
>    service dependent?)
>    2. what happens when we go from 2 nodes -> 1 node
>
> maybe it's a non-issue since we are opinionated...
>

We don't have, atm, support for 1., and we can rework it when we do. If the
user goes to the cli and does things we don't support, it's expected things
will break anyway.

As for 2., this is meant as a one time kind of thing, and to properly
support an initial single-node deployment. I think when we allow shrinking
nodes we can ask the user about that after fully informing them what is
about to happen.

tserong (dismissed):

tserong:
> blindly discarding existing rulesets that could be user defined (or service dependent?)

When we have support for those things, it should be easy enough to rework: when looping through the pools, check the current default ruleset for each, and if it's _not_ "single_node_rule", don't change it.

src/gravel/controllers/orch/ceph.py
```diff
@@ -188,6 +193,97 @@ def get_pools(self) -> List[CephOSDPoolEntryModel]:
...
+            return False
+
+        return True
+
+    def set_replicated_ruleset(self) -> bool:
+        rulesetid = self._get_ruleset_id("replicated_rule")
+        assert rulesetid >= 0
+        if not self._set_default_ruleset_config(rulesetid):
+            return False
+
```
tserong, line 264:
```suggestion

        single_node_rulesetid = self._get_ruleset_id("single_node_rule")
```

src/gravel/controllers/orch/ceph.py
```diff
@@ -188,6 +193,97 @@ def get_pools(self) -> List[CephOSDPoolEntryModel]:
...
+        pools: List[CephOSDPoolEntryModel] = self.get_pools()
+        for pool in pools:
```
tserong, line 266:
```suggestion
        for pool in pools:
            if pool.crush_rule != single_node_rulesetid: continue
```

tserong (commented):
Having thought about @mgfritch's comment a bit more, here's a completely un-tested fix :-)

tserong:
> I think when we allow shrinking nodes we can ask the user about that after fully informing them what is about to happen.

Agreed

tserong (approved):
LGTM, with one further thought - if I'm reading this right, then `set_replicated_ruleset()` will be called for every node addition, but it really only needs to happen when adding the second node (not the third, fourth, fifth, ...)

jecluis:
> LGTM, with one further thought - if I'm reading this right, then `set_replicated_ruleset()` will be called for every node addition, but it really only needs to happen when adding the second node (not the third, fourth, fifth, ...)

this is correct. But at the moment we don't have a reliable way to know that. Let me open an issue to track this.

PR #353, opened 2021-03-29
gravel/glass: present constraints on service creation

### LINKED ISSUE #86: gravel: figure out resource limitations for service creation on user input
When initiating the deployment of a service in the dashboard, ensure that the user is able to identify resource limitations. Consider:

- What resource limitations we have for a single and multi-node deploy
- How to alert the user when the resource limitation has been hit
- Identifying a resource buffer 
- If applicable, how to gain more resources

jecluis:
With this patchset we are now able to present certain constraints that are not able to be met when creating a service; e.g., number of hosts doesn't allow for availability in case of failures, or number of replicas cannot be met.

Unfortunately, my skills at showing those constraints on the UI came up short and, to be perfectly frank, it's quite ugly. Hopefully someone else will be able to make them more digestible once this is merged :)

Resolves #86 

### DIFF
--- a/src/gravel/controllers/services.py
+++ b/src/gravel/controllers/services.py
@@ -14,8 +14,11 @@
 from enum import Enum
 from pathlib import Path
 from typing import Dict, List, Tuple
+from logging import Logger
+from fastapi.logger import logger as fastapi_logger
 from pydantic import BaseModel
 from pydantic.fields import Field
+
 from gravel.controllers.orch.ceph import Mon
 from gravel.controllers.orch.cephfs import CephFS, CephFSError
 from gravel.controllers.orch.models import (
@@ -23,10 +26,18 @@ from gravel.controllers.orch.models import (
     CephOSDPoolEntryModel
 )
 from gravel.controllers.gstate import gstate
+from gravel.controllers.resources.devices import (
+    DeviceHostModel,
+    Devices,
+    get_devices
+)
 from gravel.controllers.resources.storage import (
     Storage, StoragePoolModel,
     get_storage
 )
+
+
+logger: Logger = fastapi_logger
 
 
 class ServiceError(Exception):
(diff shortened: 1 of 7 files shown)


src/gravel/controllers/services.py
```diff
@@ -152,6 +185,58 @@ def get(self, name: str) -> ServiceModel:
...
+        hosts: List[str] = \
+            [h for h, d in devs_per_host.items() if len(d.devices) > 0]
+        num_hosts: int = len(hosts)
+
+        availability: AvailabilityConstraints = \
+            AvailabilityConstraints(hosts=num_hosts)
+
+        max_devs: int = 0
+        min_devs: int = -1
+        for h in hosts:
```
jhesketh, line 209: (nit) instead of constructing hosts on line 200, the AvailabilityConstraints and appropriate min/max parts could be done in the same loop.

jhesketh (approved):
I haven't reviewed the web stuff super close, but it otherwise lgtm.

votdev:
Bildschirmfoto vom 2021-03-30 13-57-04

IMO the upper notification/warning should use a red color. The next question is: Is this only a warning or should it prevent the form from being submitted?

jecluis:
> Bildschirmfoto vom 2021-03-30 13-57-04
> 
> IMO the upper notification/warning should use a red color. The next question is: Is this only a warning or should it prevent the form from being submitted?

I don't think this should prevent the user from proceeding. Just as a way to inform them. As for color, I thought about something like orange or something, but wasn't sure what to go with so I went with nothing ;) Red just feels too error-ish for what is meant as a warning :shrug: 

> 
> The warning does not disappear when i set the number of replica to 1. Is this by intention?

Yes. Regardless of number of replicas, I thought it was important to make sure the user knows that with just one node there is no failure tolerance in terms of availability. When we add a second node that warning goes away.

jecluis:
@votdev what about this:

image

votdev (approved):

PR #354, opened 2021-03-30
Rewrite the kiwi without MicroOS and remove non essential parts

AvengerMoJo:
Same as #316 just remove all the commits by a new branch instead.
Removing dependency from MicroOS
Create LiveISO image type with persistent storage 
Updated vagrant image and add VirtualBox support. 

Signed-off-by: Alex Lau (AvengerMoJo) <alau@suse.com>


ml8mr:
@AvengerMoJo I just tried this (using your branch) and the boot process left me at a login prompt - with no sign of aquarium starting.

AvengerMoJo:
@ml8mr which image type are you using?

ml8mr:
> @ml8mr which image type are you using?

I used the live image called project-aquarium-microOS.x86_64-16.0.0.iso. (I was multi-tasking so if there was another image I should have used I missed it.)

AvengerMoJo:
> I used the live image called project-aquarium-microOS.x86_64-16.0.0.iso. (I was multi-tasking so if there was another image I should have used I missed it.)

I get the following log from my [https://paste.opensuse.org/8988607](url)

I believe the bootstrap is not being started.

AvengerMoJo:
> > I used the live image called project-aquarium-microOS.x86_64-16.0.0.iso. (I was multi-tasking so if there was another image I should have used I missed it.)
> 
> I get the following log from my [https://paste.opensuse.org/8988607](url)
> 
> I believe the bootstrap is not being started.
@ml8mr after I wait a little longer and restart the aquarium service.( not sure which is the required step) the bootstrap process able to start correctly. 
chrome-capture (15)

jecluis:
> @AvengerMoJo I just tried this (using your branch) and the boot process left me at a login prompt - with no sign of aquarium starting.

Aquarium starts as a web service, not on the node's terminal, if that's what you meant.

images/microOS/config.xml
```diff
@@ -560,31 +117,21 @@
...
-                <vmdisk id="0" controller="ide"/>
-                <vmnic driver="e1000" interface="0" mode="bridged"/>
-            </machine>
-        </type>
+            kernelcmdline="quiet systemd.show_status=yes console=ttyS0,115200 console=tty0"
+            devicepersistency="by-uuid"/>
     </preferences>
+    <users>
+         <user password="suse1234" home="/root" name="root" groups="root" pwdformat="plain"/>
+	 <user password="vagrant" home="/home/vagrant" name="vagrant" groups="vagrant" pwdformat="plain"/>
```
jecluis, line 128: is there a reason why we're adding the vagrant user here instead of doing that on the `config.sh` script along with other vagrant-related operations?

images/microOS/config.xml
```diff
@@ -560,31 +117,21 @@
...
-            <machine memory="512" guestOS="suse" HWversion="4">
-                <vmdisk id="0" controller="ide"/>
-                <vmnic driver="e1000" interface="0" mode="bridged"/>
-            </machine>
-        </type>
+            kernelcmdline="quiet systemd.show_status=yes console=ttyS0,115200 console=tty0"
+            devicepersistency="by-uuid"/>
     </preferences>
+    <users>
+         <user password="suse1234" home="/root" name="root" groups="root" pwdformat="plain"/>
```
jecluis, line 127: I'd prefer if we were to keep solely the `aquarium` user here with a default password, and add it to the sudoers list in `config.sh`, instead of adding a default `root` password. That way we'd have less stuff to ask the user to change on first run.

images/microOS/config.xml
jecluis: Or in the Vagrant preferences.

jecluis (commented):
For some reason my machine is getting stuck on boot with
```
dracut-initqueue: mount: /run/overlay/overlayfs: can't find LABEL="cow"
dracut-initqueue: The device contains 'iso9660' signature and it may remain on the device. It is recommended to wipe the device with wipefs(8) or fdisk --wipe, in order to avoid possible collisions.
dracut-initqueue: mke2fs 1.46.2 (28-Feb-2021)
```

And that's it. Stuck forever.

Other than this, the config looks sound. I'll try to figure out what's wrong, and approve once I get it working.

images/microOS/config.xml
```diff
@@ -593,120 +140,67 @@
...
+    <packages type="image" profiles="Vagrant-x86_64">
         <!-- Needed for initial configuration, except on OpenStack-Cloud, where cloud-init is used -->
         <package name="ignition-dracut"/>
         <package name="combustion"/>
         <package name="growpart-generator"/>
     </packages>
-    <packages type="image" profiles="ContainerHost">
-        <namedCollection name="container_runtime"/>
-        <package name="patterns-containers-container_runtime"/>
-        <namedCollection name="microos_apparmor"/>
```
jecluis, line 623: I think we need to include `patterns-base-apparmor`, otherwise we end up with an empty `/sys/kernel/security/apparmor/profiles`.

jecluis (changes requested):
Pretty sure we need the apparmor stuff included, but others might have a more informed opinion than _things break if not installed_.

Additionally, it feels like the image is relying on ram for its disk, instead of a persistent partition on the drive. Unsure this is due to how I created the stick, using `dd if=foo.iso of=/dev/myusbdev`.

Reality is that the disk, even though it has the two partitions that are expected (root and a ginormous partition at the end), the root fs ends up with very little space available.

```
[ceph: root@localhost /]# df -h
Filesystem      Size  Used Avail Use% Mounted on
LiveOS_rootfs   7.8G  7.8G   27M 100% /
tmpfs           7.8G     0  7.8G   0% /sys/fs/cgroup
devtmpfs        7.8G     0  7.8G   0% /dev
tmpfs           7.8G   84K  7.8G   1% /dev/shm
tmpfs           3.2G   43M  3.1G   2% /run/lvm
LiveOS_rootfs   7.8G  7.8G   27M 100% /dev/init
LiveOS_rootfs   7.8G  7.8G   27M 100% /etc/ceph/ceph.conf
```

And this is a 128GB usb stick :)

jecluis:
> For some reason my machine is getting stuck on boot with
> 
> ```
> dracut-initqueue: mount: /run/overlay/overlayfs: can't find LABEL="cow"
> dracut-initqueue: The device contains 'iso9660' signature and it may remain on the device. It is recommended to wipe the device with wipefs(8) or fdisk --wipe, in order to avoid possible collisions.
> dracut-initqueue: mke2fs 1.46.2 (28-Feb-2021)
> ```

FWIW, looks like this only happens on the first run. Upon rebooting it works fine. It might be that the second partition needs to be created prior to using the stick.

ml8mr:
> > @AvengerMoJo I just tried this (using your branch) and the boot process left me at a login prompt - with no sign of aquarium starting.
> 
> Aquarium starts as a web service, not on the node's terminal, if that's what you meant.

In the older version, wasn't there a message saying "Connect to Aquarium via https://[IP_ADDRESS]:8443"  (or something like that)?

jecluis:
> > > @AvengerMoJo I just tried this (using your branch) and the boot process left me at a login prompt - with no sign of aquarium starting.
> > 
> > 
> > Aquarium starts as a web service, not on the node's terminal, if that's what you meant.
> 
> In the older version, wasn't there a message saying "Connect to Aquarium via https://[IP_ADDRESS]:8443" (or something like that)?

I think the first PoC did, but that was back when we were doing the bootstrap on node start. We don't do that anymore, although we should!

jecluis:
My latest findings:

1. using ext4 as filesystem for the image makes it really slow, and first boot takes ages on the account of creating the third partition with ext4;
2. using xfs instead makes it much quicker.
3. everything is painfully slow in general, regardless of filesystem above -- installing `vim` on the machine is taking ages now, and I'm pretty sure it's a bandwidth problem with the drive (drive is USB3.0, but port is only USB2.0, so :shrug: )
4. podman doesn't seem to play well with the overlayfs being used, and gets stuck creating containers we _need_ to properly run Aquarium.

Regardless, running on real hardware managed to surface a couple of bugs that shall be fixed soon, that otherwise would likely not have been caught.

I'm also wondering if this might have anything to do with running on a 128G usb stick, but I find that hard to believe.

jecluis:
Let's kick this down to next week. I'll test the image more thoroughly once I get back home and have physical access to my lab.

images/microOS/config.xml
AvengerMoJo: > is there a reason why we're adding the vagrant user here instead of doing that on the `config.sh` script along with other vagrant-related operations?

I will remove that for non-vagrant image and put that into build script.

images/microOS/config.xml
AvengerMoJo: ok I will remove the root later. I keep that there for debugging in the beginning

images/microOS/config.xml
AvengerMoJo: Let me test that later with the apparmor patterns included

AvengerMoJo:
> > For some reason my machine is getting stuck on boot with
> > ```
> > dracut-initqueue: mount: /run/overlay/overlayfs: can't find LABEL="cow"
> > dracut-initqueue: The device contains 'iso9660' signature and it may remain on the device. It is recommended to wipe the device with wipefs(8) or fdisk --wipe, in order to avoid possible collisions.
> > dracut-initqueue: mke2fs 1.46.2 (28-Feb-2021)
> > ```
> 
> FWIW, looks like this only happens on the first run. Upon rebooting it works fine. It might be that the second partition needs to be created prior to using the stick.

Correct, after the first boot an expended partition will be created in your USB and mount overlay to the read only iso. So all the configuration and data will be persistent

AvengerMoJo:
> Connect to Aquarium via https://[IP_ADDRESS]:8443"

I can put something there with connecting to vagrant setup with http://localhost:1334 
something like that.

AvengerMoJo:
> In the older version, wasn't there a message saying "Connect to Aquarium via https://[IP_ADDRESS]:8443" (or something like that)?

@ml8mr I can put something there with connecting to vagrant setup with http://localhost:1334
something like that.

AvengerMoJo:
> My latest findings:
> 
> 1. using ext4 as filesystem for the image makes it really slow, and first boot takes ages on the account of creating the third partition with ext4;
> 2. using xfs instead makes it much quicker.
> 3. everything is painfully slow in general, regardless of filesystem above -- installing `vim` on the machine is taking ages now, and I'm pretty sure it's a bandwidth problem with the drive (drive is USB3.0, but port is only USB2.0, so 🤷 )
> 4. podman doesn't seem to play well with the overlayfs being used, and gets stuck creating containers we _need_ to properly run Aquarium.
> 
> Regardless, running on real hardware managed to surface a couple of bugs that shall be fixed soon, that otherwise would likely not have been caught.
> 
> I'm also wondering if this might have anything to do with running on a 128G usb stick, but I find that hard to believe.

I can update the filesystem for sure. 

Correct when I'm using USB stick vs USB SSD. SSD is working and stick fail. Both are using VM. 
I believe the podman write request is too much for stick to handle.

jecluis:
> > My latest findings:
> > 
> > 1. using ext4 as filesystem for the image makes it really slow, and first boot takes ages on the account of creating the third partition with ext4;
> > 2. using xfs instead makes it much quicker.
> > 3. everything is painfully slow in general, regardless of filesystem above -- installing `vim` on the machine is taking ages now, and I'm pretty sure it's a bandwidth problem with the drive (drive is USB3.0, but port is only USB2.0, so shrug )
> > 4. podman doesn't seem to play well with the overlayfs being used, and gets stuck creating containers we _need_ to properly run Aquarium.
> > 
> > Regardless, running on real hardware managed to surface a couple of bugs that shall be fixed soon, that otherwise would likely not have been caught.
> > I'm also wondering if this might have anything to do with running on a 128G usb stick, but I find that hard to believe.
> 
> I can update the filesystem for sure.
> 
> Correct when I'm using USB stick vs USB SSD. SSD is working and stick fail. Both are using VM.
> I believe the podman write request is too much for stick to handle.

@tserong this has me thinking that we can't get away without writing state to a partition, with or without a live persistent image, which might make the persistence on the stick a bit moot. Should we consider an install instead, or does live + mounts still appeal to you more?

AvengerMoJo:
> > I can update the filesystem for sure.
> > Correct when I'm using USB stick vs USB SSD. SSD is working and stick fail. Both are using VM.
> > I believe the podman write request is too much for stick to handle.
> 
> @tserong this has me thinking that we can't get away without writing state to a partition, with or without a live persistent image, which might make the persistence on the stick a bit moot. Should we consider an install instead, or does live + mounts still appeal to you more?

@jecluis let me also test USB stick with physical node and see the result as well.

images/microOS/config.xml
AvengerMoJo: fixed in the last commit and vagrant user only for vagrant image

AvengerMoJo:
Replace the ext4 with xfs, adopt the aqrdev (without the 3.8 requirement for build host as well for datetime conversion error) adding virtualbox template

AvengerMoJo:
jenkins run tumbleweed

AvengerMoJo:
jenkins run leap

jecluis (changes requested):
The latest commit (14f810da6ff86ffbab39a851174182a809bee7cc) should be split in at least 2 different commits:

1. a microOS-related one, with the changes to `config.sh` and `config.xml`, which should then be squashed with the original commit;
2. the aqrdev / vagrantfile template changes.

Regardless, thanks for the great work you've been putting towards this! :)

AvengerMoJo:
jenkins test leap

AvengerMoJo:
> The latest commit ([14f810d](https://github.com/aquarist-labs/aquarium/commit/14f810da6ff86ffbab39a851174182a809bee7cc)) should be split in at least 2 different commits:
> 
> 1. a microOS-related one, with the changes to `config.sh` and `config.xml`, which should then be squashed with the original commit;
> 2. the aqrdev / vagrantfile template changes.
> 
> Regardless, thanks for the great work you've been putting towards this! :)

Sure let me try to figure out how to split that, so the idea is rollback and maybe push to another PR?

jecluis:
> > The latest commit ([14f810d](https://github.com/aquarist-labs/aquarium/commit/14f810da6ff86ffbab39a851174182a809bee7cc)) should be split in at least 2 different commits:
> > 
> > 1. a microOS-related one, with the changes to `config.sh` and `config.xml`, which should then be squashed with the original commit;
> > 2. the aqrdev / vagrantfile template changes.
> > 
> > Regardless, thanks for the great work you've been putting towards this! :)
> 
> Sure let me try to figure out how to split that, so the idea is rollback and maybe push to another PR?

Not at all. We can keep this PR open. You just need to perform repush to this branch once you're done tidying it up. Let me know if you'd like some help with the squashing. :)

AvengerMoJo:
@kshtsk I'm still don't know how to resolve the jenkins error. Please advise, thanks

kshtsk:
First of all the PR have conflicts and we need to rebase it.

kshtsk:
After that we can see if it still an issue.

kshtsk:
jenkins test tumbleweed

kshtsk:
jenkins test tumbleweed

AvengerMoJo:
> jenkins test tumbleweed

I'm not sure why it is fail since  I don't have libvirt setup in my enviornment but virtualbox is working. Maybe someone with libvirt can help me debug this one?

jecluis:
From Jenkins, failure happens when the VM is trying to obtain an IP address. 

```
>>> ==> node01: Creating shared folders metadata...
>>> ==> node01: Starting domain.
>>> ==> node01: Waiting for domain to get an IP address...
EEE ERROR:aquarium:unable to start vagrant machines: /usr/lib64/ruby/gems/2.7.0/gems/fog-core-2.2.3/lib/fog/core/wait_for.rb:9:in `block in wait_for': The specified wait_for timeout (2 seconds) was exceeded (Fog::Errors::TimeoutError)
```

And I can confirm it also happens locally, except instead of throwing that exception it just gets stuck forever trying to obtain an address. I suspect this has something to do with having changed how the network devices are configured (read, eth0)

jecluis:
somehow the image is broken with libvirt:

image

jecluis:
image

For some reason the image seems broken as a root device. AFAIK, /dev/vda will be the OS boot device, which is essentially this image.

images/microOS/config.xml
```diff
@@ -1,551 +1,102 @@
...
-        <type
-            image="vmx"
-            filesystem="btrfs"
+            filesystem="xfs"
             format="vagrant"
             bootloader="grub2"
             bootloader_console="gfxterm"
             firmware="uefi"
-            kernelcmdline="quiet systemd.show_status=yes console=ttyS0,115200 console=tty0 net.ifnames=0 \$ignition_firstboot ignition.platform.id=metal"
+            kernelcmdline="quiet systemd.show_status=yes console=ttyS0,115200 console=tty0 net.ifnames=0 \$ignition_firstboot ignition.platform.id=metal root=/dev/vda1 disk=/dev/vda"
```
jecluis, line 62: this line is the culprit, and why it doesn't work with vagrant-libvirt. Please drop the ` root=/dev/vda1 disk=/dev/vda` at the end.

jecluis (changes requested):

AvengerMoJo:
> From Jenkins, failure happens when the VM is trying to obtain an IP address.
> 
> ```
> >>> ==> node01: Creating shared folders metadata...
> >>> ==> node01: Starting domain.
> >>> ==> node01: Waiting for domain to get an IP address...
> EEE ERROR:aquarium:unable to start vagrant machines: /usr/lib64/ruby/gems/2.7.0/gems/fog-core-2.2.3/lib/fog/core/wait_for.rb:9:in `block in wait_for': The specified wait_for timeout (2 seconds) was exceeded (Fog::Errors::TimeoutError)
> ```
> 
> And I can confirm it also happens locally, except instead of throwing that exception it just gets stuck forever trying to obtain an address. I suspect this has something to do with having changed how the network devices are configured (read, eth0)

Interesting if eth0 doesn't work I move it back to before using lan0 install. I don't fully understand why libvirt fails on a default eth0 network interface but I will try to see what happens.

images/microOS/config.xml
AvengerMoJo: this is also interesting .. because I don't think I touch this line and not sure why it is failing now. It should be the same the image before. But I drop them and see it helps.

AvengerMoJo:
> > And I can confirm it also happens locally, except instead of throwing that exception it just gets stuck forever trying to obtain an address. I suspect this has something to do with having changed how the network devices are configured (read, eth0)
> 
> Interesting if eth0 doesn't work I move it back to before using lan0 install. I don't fully understand why libvirt fails on a default eth0 network interface but I will try to see what happens.

I see what's wrong now, the default libvirt network card started with a random interface and when I use a persistent interface eth0 or lan0 it couldn't bind the port it is expecting. I guess I will remove eth0 or lan0 for vagrant then.

jecluis:
It had nothing to do with the network interface; eth0 was fine, and IMO more obvious what it is instead of lan0. The problem was with the vda root thing.

jecluis:
> It had nothing to do with the network interface; eth0 was fine, and IMO more obvious what it is instead of lan0. The problem was with the vda root thing.

And I mean, that was what fixed for me. :shrug:

AvengerMoJo:
> It had nothing to do with the network interface; eth0 was fine, and IMO more obvious what it is instead of lan0. The problem was with the vda root thing.

interesting ... as I recall my virtual box image have the same problem first with this image the root is actually at sda3 instead of sda1 so I assume it affected the vda1 to vda3 as well. Let me check that and see if I can figure out why the OEM image switch root to the third parttion.

AvengerMoJo:
jenkins test leap

AvengerMoJo:
OK now @jecluis  you can test the image and see if it is running correctly. Thanks

jecluis:
@AvengerMoJo it is working, but `journalctl` shows a bunch of errors regarding random things trying to write to read-only filesystems, but this one in particular seems troublesome:

```
Apr 17 20:40:50 node01 systemd-hostnamed[3883]: Failed to write static hostname: Permission denied
```

The hostname was nevertheless changed. I'm guessing it is failing to write it down somewhere though.

Besides that, Aquarium seems to be working just fine.

jecluis:
Vagrant image currently breaks when running with 2 or more network interfaces.

```
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

/sbin/ifdown 'eth1' || true
mv '/tmp/vagrant-network-eth1-1618773225-0' '/etc/sysconfig/network/ifcfg-eth1'
/sbin/ifup 'eth1'


Stdout from the command:

eth1            no-device


Stderr from the command:

wicked: ifdown: no matching interface
```

This does not happen with the MicroOS image we currently have.

AvengerMoJo:
I can see the different now. One major different is the old image does see the new interface but I also get a fail with vagrant 

The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

/sbin/ifdown 'eth1' || true
mv '/tmp/vagrant-network-eth1-1618972267-0' '/etc/sysconfig/network/ifcfg-eth1'
/sbin/ifup 'eth1'


Stdout from the command:

eth1            setup-in-progress


Stderr from the command:

wicked: skipping eth1 interface: device is not configured by wicked yet
wicked: ifdown: no matching interfaces

AvengerMoJo:
in the old image, 

cat /etc/sysconfig/network/ifcfg-eth1

#VAGRANT-BEGIN
# The contents below are automatically generated by Vagrant. Do not modify.
BOOTPROTO='dhcp'
STARTMODE='auto'
DEVICE='eth1'
#VAGRANT-END

So it is exactly the last step of settiing new interface up and moving the configuration by vagrant. I still trying to figure out why it is not working in my image but it should be ok.

kshtsk:
rebase and resolve conflicts are required

tserong:
Just for the record, we found the problem with multi-nic vagrant deployments - the images/microOS/root/etc/udev/rules.d/70-persistent-net.rules file is what messes things up. If we remove that, vagrant works just fine again. (I wonder if we should also remove images/microOS/root/usr/lib/systemd/network/80-mac-net.link, as that wasn't present before either? That one at least didn't seem to cause me any trouble with vagrant though)

tests/vagrant/templates/Vagrantfile.tmpl
```diff
@@ -0,0 +1,40 @@
+# -*- mode: ruby -*-
```
tserong, line 1: tests/vagrant/templates/Vagrantfile.tmpl should be removed, as it's not used anymore (Vagrantfiles are generated automatically by `aqrdev` now)

AvengerMoJo:
jenkins run leap

tserong:
jenkins run tumbleweed

tserong:
(looks like the tumbleweed run Alex triggered yesterday has been stuck for, like, a day on the ">>> + git fetch --tags --progress https://github.com/aquarist-labs/aquarium '+refs/heads/*:refs/remotes/origin/*'" step, so I'm hoping asking for a re-run will unstick it....

tserong (approved):
I've just re-tested the vagrant build with 2 NICs, and it works fine. As discussed with @AvengerMoJo on Slack, further improvements can happen in another branch / subsequent PRs, so I reckon we ought to merge this one now. Any objections @jecluis ?

kshtsk:
jenkins test tumbleweed

kshtsk:
jenkins test tumbleweed

PR #355, opened 2021-03-30
glass: round size values to 2 decimals if needed

jecluis:
image


s0nea (approved):
Thanks for the enhancement :+1:

PR #358, opened 2021-03-30
gravel: endpoints to manage nfs exports

### LINKED ISSUE #81: gravel: nfs support: creation via orchestrator interface
Ensure we have support for the NFS client and operational management via the orchestartor interface. Consider:

- How to deploy NFS/bootstrap configuration
- Starting and restarting nfs (in the dashboard?)
- How to manage NFS exports
- Using CephFS or RGW as the backstore
- Using the dashboard to manage NFS ganesha config files
- Configuring multiple NFS ganesha clusters

mgfritch:
resolves #81.

### DIFF
--- a/src/gravel/api/nfs.py
+++ b/src/gravel/api/nfs.py
@@ -21,7 +21,8 @@ from fastapi.routing import APIRouter
 from pydantic import BaseModel
 
 from gravel.controllers.orch.nfs import \
-    NFSError, NFSService, NFSServiceModel
+    NFSError, NFSExport, NFSExportModel, \
+    NFSService, NFSServiceModel
 
 
 logger: Logger = fastapi_logger
@@ -100,3 +101,21 @@ def get_service_info(service_id: str) -> NFSServiceModel:
     except NFSError as e:
         raise HTTPException(status.HTTP_428_PRECONDITION_REQUIRED,
                             detail=str(e))
+
+
+@router.get(
+    '/export/{service_id}/{export_id}',
+    name='nfs export detail',
+    response_model=NFSExportModel)
+async def get_export_info(service_id: str, export_id: int) -> NFSExportModel:
+    try:
+        res = NFSExport().info(service_id)
+    except NFSError as e:
+        raise HTTPException(status.HTTP_428_PRECONDITION_REQUIRED,
+                            detail=str(e))
+
+    for export in res:
+        if export.export_id == export_id:
+            return export
+
+    raise HTTPException(status.HTTP_404_NOT_FOUND)
--- a/src/gravel/api/nfs.py
+++ b/src/gravel/api/nfs.py
@@ -41,12 +41,12 @@ class Response(BaseModel):
 
 
 @router.put(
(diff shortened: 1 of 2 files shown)


src/gravel/api/nfs.py
```diff
@@ -103,6 +112,28 @@ def get_service_info(service_id: str) -> NFSServiceModel:
                             detail=str(e))
 
 
+@router.post(
+    '/export/{service_id}',
```
tserong, line 116: This is where you wanted to be able to do something like`@router.put('/export/{service_id}/{path}', ...)`, but couldn't because path is a path, and messes up the URL, right? Can you use https://fastapi.tiangolo.com/tutorial/path-params/#path-convertor here?

src/gravel/controllers/orch/nfs.py
```diff
@@ -78,23 +100,100 @@ def ls(self) -> List[str]:
...
+            'format': 'json',
+        }
+
+        res = self._call(cmd)
+
+        ret: List[NFSExportModel] = []
+        for export in res:
+            ret.append(
+                NFSExportModel(
+                    **export
```
tserong, line 196: What happens if the data returned from `ceph nfs export ls` changes in future (extra fields, or whatever)? Does this still do the right thing? (Answer, having done a little further poking around: probably, yes, because pydantic should only pull in data that matches the model)

mgfritch: yes, the extra args are ignored if they are not defined by the model. However, if the model defines an arg that does not exist, pydantic will raise an error.

src/gravel/api/nfs.py
mgfritch: Yes! So the question would be how do we prefer to represent the endpoints?

Using the export id:
`GET /nfs/export/<int>`

Using the psudo path:
`GET /nfs/export/foo/bar/baz`

Maybe we rework this in a follow-up PR?

jecluis: I think passing the path in the URL is bound to cause confusion. Passing it as a request is cleaner and tidier. :shrug:

src/gravel/controllers/orch/nfs.py
```diff
@@ -26,15 +28,33 @@ class NFSDaemonModel(BaseModel):
...
+class NFSExportModel(BaseModel):
+    export_id: int
+    path: str
+    pseudo: str
+    access_type: str
+    squash: str
+    security_label: bool
+    protocols: List[str]
+    transports: List[str]
+    fsal: Dict  # TODO: create model for this?
```
jecluis, line 49: nit: this is likely going to send pyright into throwing a fit because of missing arguments to `Dict`. `Dict[Any, Any]` would be preferable.

jecluis (approved):
Haven't tested the code, but it seems sound. I now realize we merged the previous PR without sanity checks as to whether there are name collisions on service creation, other than the expected error we'll likely get from calling on create for an existing service; but surely this can be addressed later on.

Issue #82, opened 2021-02-10
gravel: nfs support - expose through rest API


jecluis:
The API should expose the options to manage NFS. Consider:

- export name
- expected space usage (so we can keep track of requirements)
- redundancy level

At this point, I think it's safe to ignore authentication.

asettle:
@mgfritch update?

PR #366, opened 2021-03-31
gravel: ctrl/services: add NFS service create

mgfritch:
create a cephfs, nfs service, and nfs export under a single svc

Resolves: #82

### DIFF
--- a/src/gravel/controllers/services.py
+++ b/src/gravel/controllers/services.py
@@ -25,6 +25,12 @@ from gravel.controllers.orch.models import (
     CephFSListEntryModel,
     CephOSDPoolEntryModel
 )
+from gravel.controllers.orch.nfs import (
+    NFSBackingStoreEnum,
+    NFSError,
+    NFSExport,
+    NFSService
+)
 from gravel.controllers.gstate import gstate
 from gravel.controllers.resources.devices import (
     DeviceHostModel,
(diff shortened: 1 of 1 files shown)


mgfritch:
Blocked by: #358

src/gravel/controllers/services.py
```diff
@@ -332,6 +340,28 @@ def get_pool(name: str) -> CephOSDPoolEntryModel:
...
             # client.
 
+    def _create_nfs(self, svc: ServiceModel) -> None:
+        # create a cephfs
+        self._create_cephfs(svc)
+
+        # create an generic NFS service
+        nfs_svc_id = 'gravel'
+        try:
+            NFSService().create(nfs_svc_id, placement='*')
```
jecluis, line 350: does this mean we only get to have one NFS service?

mgfritch: yes, one NFS service an NFS export per cephfs.

co-location of NFS services is wip: https://github.com/ceph/ceph/pull/40543

jecluis: Okay, so, I understand this might seem intuitive but I had to go run the commands myself to ensure the creation command is in fact idempotent. Mind adding a note there stating that this will indeed call `nfs cluster create` multiple times, once per service creation, but that the underlying command is idempotent as thus we get away with it?

I know this feels like a little thing, but from the code, as is, really seems like we could just be bailing out through an exception should the service already exist.

mgfritch: actually, good catch! I think your intuition is right .. 

The command appears to be idempotent, but maybe it should not be, because if the supplied placement spec happens to differ, this is not currently updated via the `create` command. Likewise, there is also no guarantee the current behavior won't change out from under us later.

I've fixed up the commit to first look for an existing cluster and update the placement or create a new cluster if one does not exist.

jhesketh (approved):
lgtm, but I haven't tested it.

PR #367, opened 2021-04-01
gravel: cephadm: lvs may not have osd info

jecluis:
If the LV is not a ceph osd, then we won't have the fields we'd be
expecting out of an osd partition.


jecluis:
glass unit tests failure is unrelated to this patchset.

mgfritch (approved):

Issue #235, opened 2021-03-01
gravel: propagate exceptions' errors properly and clearly during service creation


jecluis:
Right now we are propagating exceptions in such a way that a proper message does not end up being conveyed through the API, nor are we actually logging anything worthwhile.

We need to be logging a lot more output coming from Ceph whenever we get bit by an error, and we need to propagate that through exceptions to the layers above us. Otherwise it is essentially maddening trying to debug any kind of misbehavior.

jecluis:
@mgfritch is this something achievable for M2, or should we push it further?

PR #369, opened 2021-04-01
aqrdev: add 'restart' command

jecluis:

### DIFF
--- a/tools/aqrdev
+++ b/tools/aqrdev
@@ -821,6 +821,21 @@ def cmd_stop(ctx: AppCtx, name: str) -> None:
         sys.exit(rc)
 
 
+@app.command("restart")
+@click.argument("name", required=True, type=str)
+@pass_appctx
+def cmd_restart(ctx: AppCtx, name: str) -> None:
+    _cmd_try_vagrant_deployment(name)
+    
+    rc: int = _stop(name)
+    if rc != 0:
+        sys.exit(rc)
+    rc = _start(name, False)
+    if rc != 0:
+        sys.exit(rc)
+
+
+
 @app.command("shell")
 @click.argument("name", required=True, type=str)
 @click.option("-n", "--node", required=False, type=str)
--- a/tools/aqrdev
+++ b/tools/aqrdev
@@ -83,6 +83,10 @@ class DeploymentStatusError(BaseError):
     pass
 
 
+class DeploymentStateError(BaseError):
+    pass
+
+
 class DeploymentModel(BaseModel):
     name: str
     created_on: str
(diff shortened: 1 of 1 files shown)


AvengerMoJo:
Not exactly related to this commit but I have two comment with aqrdev
1) the aqrdev require multiple pip install for the host may need to put into the setup.sh 
2) the doc also need to be updated accordingly.

jecluis:
I don't know what you mean. All the dependencies for aqrdev (read, click) are on requirements.txt and are installed into the venv automatically, and the docs were updated accordingly to use aqrdev for what matters for the docs. What is it that is missing?

AvengerMoJo:
> I don't know what you mean. All the dependencies for aqrdev (read, click) are on requirements.txt and are installed into the venv automatically, and the docs were updated accordingly to use aqrdev for what matters for the docs. What is it that is missing?

sorry about that I was testing just the deployment different machines and forget to run the setup-dev.sh nevermind ..

tools/aqrdev
```diff
@@ -620,11 +630,16 @@ def _get_deployment_status(name: str) -> Dict[str, List[Tuple[str, str]]]:
 
     path: Path = _get_deployment_path(name)
     if not path.exists():
-        raise DeploymentPathNotFoundError()
+        raise DeploymentPathNotFoundError(
+            "deployment not found",
+            rc=errno.ENOENT
```
mgfritch, line 635:
```suggestion
            rc=-errno.ENOENT
```

use negative errno int?

tools/aqrdev
```diff
@@ -620,11 +630,16 @@ def _get_deployment_status(name: str) -> Dict[str, List[Tuple[str, str]]]:
...
+        )
 
     vagrantfile = path.joinpath("Vagrantfile")
     if not vagrantfile:
-        raise DeploymentNotFinishedError()
+        raise DeploymentNotFinishedError(
+            "deployment not finished", rc=errno.ENOENT
```
mgfritch, line 641: I would almost prefer to reverse these args (e.g. errno appear first), but we'd need to extend the Error class todo that.

wdyt if I were to rework this via a PR #235 ?

mgfritch (approved):
couple nits, but otherwise lgtm!

tools/aqrdev
jecluis: > wdyt if I were to rework this via a PR #235 ?

sounds good to me!

jhesketh:
So the documentation is a little light on, but should we be using ClickException for our base exception handler (https://click.palletsprojects.com/en/7.x/exceptions/#where-are-errors-handled). This has support for `exit_code`'s which can be appropriately bubbled through click's sys.exit (https://click.palletsprojects.com/en/7.x/commands/#command-return-values).

jecluis:
dropping this on the floor as a major refactor is incoming.

PR #372, opened 2021-04-06
glass: Rename CSS class name

votdev:

### DIFF
--- a/src/glass/src/styles.scss
+++ b/src/glass/src/styles.scss
@@ -81,7 +81,7 @@ $glass-color-accent-complementary: white;
   height: 48px;
   width: 48px;
 }
-.glass-icon-large {
+.glass-icon-4x {
   height: 96px;
   width: 96px;
 }


s0nea (approved):
I'm wondering if we need the class if it isn't used at all. Or do you think we will need it in the future? - Anyway... the change looks sane to me.

votdev:
> I'm wondering if we need the class if it isn't used at all. Or do you think we will need it in the future? - Anyway... the change looks sane to me.

This class is already used at different places, but it seems one of the PR that renamed the class has missed to do it in the `styles.scss` file.

s0nea:
> > I'm wondering if we need the class if it isn't used at all. Or do you think we will need it in the future? - Anyway... the change looks sane to me.
> 
> This class is already used at different places, but it seems one of the PR that renamed the class has missed to do it in the `styles.scss` file.

Ah... I see. Thanks!

PR #374, opened 2021-04-06
gravel: fix ctrl/status and deduplicate some code

jecluis:
We introduced a regression a while ago while fixing a bug, making the status controller tick while a node is joining, which should not happen because, at the time, we don't have a ceph.conf. This patchset fixes this.

Additionally, we're deduplicating some code in the ceph controller.

### DIFF
--- a/src/gravel/controllers/orch/ceph.py
+++ b/src/gravel/controllers/orch/ceph.py
@@ -250,6 +250,35 @@ class Mon(Ceph):
             return False
         return True
 
+    def pool_set(
+        self,
+        poolname: str,
+        var: str,
+        value: str,
+        really: bool = False
+    ) -> bool:
+        """ Set given pool's configuration variable to a provided value """
+        assert poolname
+        assert var
+        assert value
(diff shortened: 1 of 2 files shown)


tserong (dismissed):

asettle (approved):

src/gravel/controllers/orch/ceph.py
```diff
@@ -234,21 +234,62 @@ def _get_ruleset_id(self, name: str) -> int:
             logger.exception(e)
             raise NoRulesetError()
 
-    def _set_default_ruleset_config(self, rulesetid: int) -> bool:
+    def config_set(self, name: str, value: str, who: str) -> bool:
```
mgfritch, line 237:
```suggestion
    def config_set(self, who: str, name: str, value: str) -> bool:
```
 (nit) but order these according to the mon command?

mgfritch (approved):
minor nit, but lgtm!

mgfritch (approved):

PR #375, opened 2021-04-06
glass: fix cephfs dialog issues

s0nea:
* glass: rename CephFS menu item: Rename the `CephFS Share` menu item into `CephFS Service` to keep consistency.
* glass: add missing translation markers in CephFS dialog
* glass: add missing license header

### DIFF
--- a/src/glass/src/app/core/modals/cephfs/cephfs-modal.component.ts
+++ b/src/glass/src/app/core/modals/cephfs/cephfs-modal.component.ts
@@ -7,6 +7,7 @@ import * as _ from 'lodash';
 import { BlockUI, NgBlockUI } from 'ng-block-ui';
 import { finalize } from 'rxjs/operators';
 
+import { translate } from '~/app/i18n.helper';
 import {
   CheckRequirementsReply,
   Constraints,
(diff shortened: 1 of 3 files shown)


src/glass/src/app/core/modals/cephfs/cephfs-modal.component.ts
```diff
@@ -79,14 +80,14 @@ export class CephfsModalComponent implements OnInit {
       return;
     }
     const values = this.formGroup.value;
-    this.blockUI.start(TEXT('Please wait, assessing service specifications ...'));
+    this.blockUI.start(translate(TEXT('Please wait, assessing service specifications ...')));
```
Devp00l, line 97: Doesn't `TEXT` which is `marker` from `ngx-translate-extract-marker` already translate the text?

s0nea: Hmmm if we don't need it here, we might need to rework 

Screenshot_2021-04-06_13-00-22

votdev: > Doesn't `TEXT` which is `marker` from `ngx-translate-extract-marker` already translate the text?

No, it's only used for marking text to be extracted.

votdev (approved):

PR #377, opened 2021-04-06
glass: Fix dashboard widgets issue related to valueFormatting

votdev:
Do not use foo.bind(this) in combination with 'valueFormatting', otherwise Angular CD will get confused and assumes false changes.

Fix various issues in the `Performance` dashboard widget:
- Translate title text
- Add workaround to force charts to adapt their size to the parent container. Better approaches are welcome.


Devp00l (approved):
Works fine :)

PR #381, opened 2021-04-07
glass: Show token required to join new node

votdev:
- Add 'Hosts' page.
- Add button on this page to show the token that is required to join a new node.
- Add auto-reload feature to datatable component.

Peek 2021-04-07 18-47

### DIFF
--- a/src/glass/src/app/core/modals/nfs/nfs-modal.component.html
+++ b/src/glass/src/app/core/modals/nfs/nfs-modal.component.html
@@ -77,7 +77,7 @@
                     fxFlex>
         </mat-slider>
         <mat-hint>
-          <span>{{formGroup.get('replicas')?.value | redundancyLevel:'long'}}</span>
+          <span>{{ formGroup.get('replicas')?.value | redundancyLevel:'long' }}</span>
         </mat-hint>
       </mat-form-field>
       <div fxLayout="row"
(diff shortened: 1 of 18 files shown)


jecluis (dismissed):

src/glass/src/app/core/modals/nfs/nfs-modal.component.html
```diff
@@ -77,7 +77,7 @@ <h1 mat-dialog-title>NFS</h1>
                     fxFlex>
         </mat-slider>
         <mat-hint>
-          <span>{{formGroup.get('replicas')?.value | redundancyLevel:'long'}}</span>
+          <span>{{ formGroup.get('replicas')?.value | redundancyLevel:'long' }}</span>
```
s0nea, line 80: We do have the same line on the `CephfsModalComponent`: https://github.com/aquarist-labs/aquarium/blob/main/src/glass/src/app/core/modals/cephfs/cephfs-modal.component.html#L80

s0nea (approved):

src/glass/src/app/core/modals/nfs/nfs-modal.component.html
votdev: Doesn't matter. We do not have formatting rules related to this. It's my personal preference to format such strings this way. I changed that only because i came across this code while doing the implementation for this PR. Angular itself seems not to have strict rules, they use both `{{foo}}`  and `{{ foo }}`, see [this example](https://angular.io/guide/interpolation#preventing-name-collisions).

s0nea: I fully agree with you. I'm also in favor of the latter.

PR #382, opened 2021-04-08
tools: Add Makefile for building dist tarball

tserong:
I've added a Makefile with a `dist` target, so you can now run `make dist` to get a tarball named something like aquarium-0.1.0-0-g2b6942a.tar.gz, which includes a built glass and all the necessary bits of gravel.

This doesn't change the current dev workflow at all.  You still build an image by running `./tools/build-image.sh`, but that script will in turn call `make dist` to get a tarball to install in the image.

Note: the version string used in the tarball name is based on the latest annotated tag in the form 'vX.Y.Z', which means I've suddenly introduced the notion of version numbers and releases to the project.  Until such time as an actual release is tagged, the version will default to v0.0.0, just to have the build working.

Fixes: https://github.com/aquarist-labs/aquarium/issues/277

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,32 @@
+# The version needs an annotated tag in the git repo in the form vX.Y.Z.
+# If such is not present, fall back to using v0.0.0 so we get a non-empty
+# version string.
+VERSION=$(shell (git describe --match 'v*' --long 2>/dev/null || echo 'v0.0.0') | sed 's/^v//')
+
+usage:
+	@echo "Try running \`make dist\` to build a tarball"
+
+submodules:
+	git submodule update --init
+
+glass:
+	cd src/glass && npm install && npx ng build --prod --output-hashing=all
+
+dist: submodules glass
+	$(eval TMPDIR := $(shell mktemp -d))
+	$(eval TAR_BASE := $(TMPDIR)/aquarium-$(VERSION))
+	$(eval TAR_USR := $(TAR_BASE)/usr/share/aquarium)
+	$(eval TAR_UNIT := $(TAR_BASE)/usr/lib/systemd/system)
+	$(eval TAR_SBIN := $(TAR_BASE)/usr/sbin)
+	mkdir -p $(TAR_USR) $(TAR_UNIT) $(TAR_SBIN)
+	# Copy gravel, glass, aquarium.py and cephadm from src/...
+	cd src && \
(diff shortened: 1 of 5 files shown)


tserong:
jenkins run tumbleweed

Makefile
```diff
@@ -0,0 +1,32 @@
...
+VERSION=$(shell (git describe --match 'v*' --long 2>/dev/null || echo 'v0.0.0') | sed 's/^v//')
+
+usage:
+	@echo "Try running \`make dist\` to build a tarball"
+
+submodules:
+	git submodule update --init
+
+glass:
+	cd src/glass && npm install && npx ng build --prod --output-hashing=all
```
jhesketh, line 13: This is my makefile ignorance here, but wouldn't it be better to use `pushd` then `popd` to ensure that the following steps are in the expected dir (eg on line 23)?

(I figure it doesn't actually matter if this is working)

tools/build-image.sh
```diff
@@ -131,63 +131,24 @@ fi
 
 set -xe
 
-glassdir=${srcdir}/glass
-graveldir=${srcdir}/gravel
-
-[[ ! -d "${glassdir}" ]] && \
-  error_exit "missing frontend directory at '${glassdir}"
```
jhesketh, line 138: As a bit of a regression we lose some of these checks. I'm not sure of their value though; if people are deleting things from the source they can't expect it to just build.

jhesketh (approved):
The Jenkins jobs verify that this works, and it otherwise lgtm; just some nits.

I also wonder if we should use `make` as the developer entry point too in due course.

Makefile
tserong: I don't think it makes a material difference in this case. The `cd` at the start here only has effect for the duration of that line, it doesn't change the overall environment, i.e. if there were another command in that block, when it was invoked, it'd be back in the original directory automatically.

tools/build-image.sh
tserong: IMO those checks aren't especially useful; as you say, someone would have had to delete something from the source tree to hit them. If someone does that with the Makefile form, they'll hit an error anyway (e.g. `cd glass` will fail with something like `/bin/sh: line 0: cd: src/glass: No such file or directory`)

tserong:
> I also wonder if we should use `make` as the developer entry point too in due course.

Yep! This is a first step along that path :-)

PR #403, opened 2021-04-10
gravel: add etcd

### LINKED ISSUE #32: gravel: persistent state

jecluis:
We need something to share state across multiple nodes, consistently, and, ideally, without a single point of failure. This patchset proposes etcd as that thing.

## Why etcd?
Because it exists, is simple enough to setup, and does not require Ceph to work.

## Why not Ceph's monitors kvstore?
We _could_ use the monitors' config-key/kv store, but a few things strike me as good enough reasons not to:
1. That was never what it was supposed to exist for, and everyone keeps abusing it. Ceph needs something like etcd to keep its stuff, or to implement a proper kvstore may it be in the monitors or elsewhere (the manager's would be preferable tbh).
2. We lose the store if the cluster dies, or doesn't start, or a monitor fills up to the brim and causes a quorum loss. If that happens, Aquarium needs to still be available to provide info to the user, and to tell the user which services have gone away, stuff like that.
3. The ceph-mon kvstore lacks a few things that are pretty cool, like watching key updates, and transactions.

## Few things of note
- we add a new dependency, on [aetcd3](https://github.com/martyanov/aetcd3). This has been the only asyncio etcd3 library I found that seems recent enough, with decent testing, and actually building on pypi, even though it doesn't seem to be particularly active.
- aetcd3 is far from perfect, and has this one thing that bites us on quit, where resources are not properly freed and an exception is raised. I've opened a PR (https://github.com/martyanov/aetcd3/pull/4) against it but I have very little hope to get it merged. This is why we're relying, for the purposes of this patchset, on a custom-compiled library sitting in one of my servers, somewhere.
- Unless there's suddenly more activity on that lib, we may have to fork it and carry it ourselves.

Resolves #32 

### DIFF
--- a/images/microOS/config.sh
+++ b/images/microOS/config.sh
@@ -174,7 +174,7 @@ EOF
 fi
 
 if [[ "$kiwi_profiles" == *"Ceph"* ]]; then
-  pip install fastapi uvicorn
+  pip install fastapi uvicorn aetcd3
   baseInsertService aquarium
 fi
 
--- a/images/microOS/config.sh
+++ b/images/microOS/config.sh
@@ -174,7 +174,8 @@ EOF
 fi
 
 if [[ "$kiwi_profiles" == *"Ceph"* ]]; then
-  pip install fastapi uvicorn aetcd3
+  pip install fastapi uvicorn \
+    https://celeuma.wipwd.dev/aqr/aetcd3/aetcd3-0.1.0a5.dev2+g0aa852e.d20210410-py3-none-any.whl
   baseInsertService aquarium
 fi
 
--- a/src/gravel/controllers/kv.py
+++ b/src/gravel/controllers/kv.py
@@ -1,0 +1,122 @@
+# project aquarium's backend
+# Copyright (C) 2021 SUSE, LLC.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+import asyncio
+from typing import Callable, Optional
+import aetcd3
+import aetcd3.locks
+import aetcd3.events
+import grpclib.exceptions
+
+from logging import Logger
+from fastapi.logger import logger as fastapi_logger
+
+logger: Logger = fastapi_logger
+
+
+class Lock:
+    _lock: aetcd3.locks.Lock
+    _is_acquired: bool
+
+    def __init__(self, lock: aetcd3.locks.Lock):
+        self._lock = lock
+        self._is_acquired = False
+
+    async def acquire(self) -> None:
+        await self._lock.acquire()
+        self._is_acquired = True
+
+    async def release(self) -> None:
+        if not self._is_acquired:
+            return
+        await self._lock.release()
+        self._is_acquired = False
+
+
+class KV:
+
+    _client: Optional[aetcd3.Etcd3Client]
+    _is_closing: bool
+
+    def __init__(self):
+        self._client = None
+        self._is_open = False
+        self._is_closing = False
+
+    async def ensure_connection(self) -> None:
+        """ Open k/v store connection """
+        # try getting the status, loop until we make it.
+        opened = False
+        while not self._is_closing:
+            try:
+                async with aetcd3.client() as client:
+                    await client.status()
+            except Exception:
+                logger.warn("etcd not up yet? sleep.")
+                await asyncio.sleep(1.0)
+                continue
+            opened = True
+            break
+        if opened:
+            self._client = aetcd3.client()
+            logger.info("opened kvstore connection")
+
+    async def close(self) -> None:
+        """ Close k/v store connection """
+        self._is_closing = True
+        if not self._client:
+            return
+        await self._client.close()
+        self._client = None
+
+    async def put(self, key: str, value: str) -> None:
+        """ Put key/value pair """
+        assert self._client
+        await self._client.put(key, value)
+
+    async def get(self, key: str) -> str:
+        """ Get value for provided key """
+        assert self._client
+        value, _ = await self._client.get(key)
+        return value.decode("utf-8")
+
+    async def rm(self, key: str) -> None:
+        """ Remove key from store """
+        assert self._client
+        await self._client.delete(key)
+
+    async def lock(self, key: str) -> Lock:
+        """ Lock a given key. Requires compliant consumers. """
+        assert self._client
+        return Lock(self._client.lock(key))
+
+    async def watch(
+        self,
+        key: str,
+        callback: Callable[[str, str], None]
+    ) -> int:
+        """ Watch updates on a given key """
+        assert self._client
+
+        async def _cb(what: aetcd3.events.Event) -> None:
+            if not what or \
+               type(what) == grpclib.exceptions.StreamTerminatedError:
+                return
+            callback(what.key.decode("utf-8"), what.value.decode("utf-8"))
+
+        return await self._client.add_watch_callback(key, _cb)
+
+    async def cancel_watch(self, watch_id: int) -> None:
+        """ Cancel a watch """
+        assert self._client
+        await self._client.cancel_watch(watch_id)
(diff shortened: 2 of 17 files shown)


src/gravel/controllers/nodes/mgr.py
```diff
@@ -344,6 +347,53 @@ async def finish_bootstrap_cb(
...
+
+        client_url: str = f"http://{address}:2379"
+        peer_url: str = f"http://{address}:2380"
+        args_dict: Dict[str, str] = {
+            "name": hostname,
+            "initial-advertise-peer-urls": peer_url,
+            "listen-peer-urls": peer_url,
+            "listen-client-urls": f"{client_url},http://127.0.0.1:2379",
+            "advertise-client-urls": client_url,
+            "initial-cluster": f"{hostname}={peer_url}",
```
tserong: So the name of the etcd cluster is the hostname of the first node. Would it make sense to use `"initial-cluster": f"aquarium={peer_url}"` instead?

src/gravel/controllers/nodes/messages.py
```diff
@@ -41,6 +41,7 @@ class WelcomeMessageModel(BaseModel):
     pubkey: str
     cephconf: str
     keyring: str
+    peer_url: str
```
tserong, line 44: Maybe I'm just nitpicking, but can we give this a slightly different name? Because it's peer_url, it makes me think it's a URL, but it's actually a string in the form "name=URL". Maybe call it, ummm... etcd_peer? Just something to give a little more information.

images/microOS/config.sh
```diff
@@ -174,7 +174,8 @@ EOF
 fi
 
 if [[ "$kiwi_profiles" == *"Ceph"* ]]; then
-  pip install fastapi uvicorn aetcd3
+  pip install fastapi uvicorn \
+    https://celeuma.wipwd.dev/aqr/aetcd3/aetcd3-0.1.0a5.dev2+g0aa852e.d20210410-py3-none-any.whl
```
tserong, line 178: Is this referring to the fix for https://github.com/aquarist-labs/aquarium/pull/403, which was since closed? In that case, do we still need this?

src/requirements.txt
```diff
@@ -7,4 +7,4 @@ starlette==0.13.6
 uvicorn==0.13.3
 pip
 websockets==8.1
-aetcd3
+https://celeuma.wipwd.dev/aqr/aetcd3/aetcd3-0.1.0a5.dev2+g0aa852e.d20210410-py3-none-any.whl
```
tserong, line 10: Same comment as above

images/microOS/config.sh
tserong, line 178: Is this referring to the fix for https://github.com/martyanov/aetcd3/pull/4, which was since closed? In that case, do we still need this?

src/requirements.txt
tserong, line 10: Same comment as in `pip install` line

src/gravel/controllers/nodes/mgr.py
```diff
@@ -435,11 +435,9 @@ def _bootstrap_process():
...
             loop.run_until_complete(_run_etcd())
 
-        process = multiprocessing.Process(
-            target=_bootstrap_process
-        )
-        process.start()
-        logger.info(f"started etcd process pid = {process.pid}")
+        t = threading.Thread(target=_bootstrap_process)
+        t.start()
+        logger.info("started etcd thread")
```
tserong, line 440: "Starting it as a process, for some reason, allows etcd to squash the event loop's signal handlers, which breaks our "on shutdown" cleanup." Huh? As a separate process it shouldn't be able to do anything to the parent, should it? And if it can mess up the parent _process_, why wouldn't it also mess up the process when run as a thread? (I kinda feel like etcd should really be a separate process, just because it's its own daemon, rather than being some class or library that we import...)

src/gravel/controllers/kv.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def __init__(self):
+        self._client = None
+        self._is_open = False
+        self._is_closing = False
+
+    async def ensure_connection(self) -> None:
+        """ Open k/v store connection """
+        # try getting the status, loop until we make it.
+        opened = False
```
tserong, line 59: What happens if we never make it? The KV store is unavailable forever?

tserong (commented):
Looks like it should work :-) Missing a cache, for the case where etcd is dead, but you already noted that we'll need a cache in one of the commit messages, so that's fine. I have no particular reason not to approve this, but a few questions inline.

src/gravel/controllers/nodes/messages.py
tserong: (oops, that comment should have been part of the review below)

images/microOS/config.sh
jecluis: I'll update the initial comment, but no. Even though that PR was closed, there's a couple of additional patches on top of the original repo, one of them to fix an aiofiles's dependency version -- original repo requires <0.6, while uvicorn requires >= 0.6, and it's annoying to have to deal with those dependencies when pip installs. Another one is a missing `await` when shutting down the lib IIRC.

src/gravel/controllers/nodes/mgr.py
jecluis: From the docs, that seems to be the daemon's instance name. `initial-cluster` seems much like `mon_initial_cluster` from Ceph. Even if that might set the cluster's name to the name of the first node, maybe there's a way to do that without having to name the first node something that does not match the hostname? (I'll look into that)

src/gravel/controllers/kv.py
jecluis: yeah. This might warrant a timeout or something. But yeah, that's it.

src/gravel/controllers/nodes/messages.py
jecluis: yep, sounds good.

jecluis (commented):
For some reason github did not allow me to reply to @tserong without reviewing my own PR? :shrug:

src/gravel/controllers/nodes/mgr.py
jecluis: wtf - this last comment should have been a reply to something else entirely. github seems to be on fire today.

Anyhoo, to @tserong actual comment:



> "Starting it as a process, for some reason, allows etcd to squash the event loop's signal handlers, which breaks our "on shutdown" cleanup." Huh? As a separate process it shouldn't be able to do anything to the parent, should it? And if it can mess up the parent _process_, why wouldn't it also mess up the process when run as a thread? (I kinda feel like etcd should really be a separate process, just because it's its own daemon, rather than being some class or library that we import...)

Yeah, I know. This is weird, really. I assumed the same thing, but for some reason, when starting an asyncio subprocess from a multithreading.Process, the signal handler is squashed; when starting the asyncio subprocess from a threading.Thread, it does not. I'll try getting a script to reproduce the behavior just to ensure I'm not imagining things though.

src/gravel/controllers/nodes/mgr.py
```diff
@@ -493,63 +610,44 @@ def connmgr(self) -> ConnMgr:
...
-        self._state = NodeStateModel.parse_file(statefile)
+        def _watcher(key: str, value: str) -> None:
+            if key == "/nodes/token":
+                logger.info(f"got updated token: {value}")
+                self._token = value
 
-    def _load(self) -> None:
-        self._token = self._load_token()
+        self._token = await self._load_token()
+        await self._kvstore.watch("/nodes/token", _watcher)
```
jhesketh, line 627: Why is a token obtained, and then the key watched for updates? Do we expect the token to change?

src/gravel/controllers/kv.py
tserong: It's probably fine for now. It'll also presumably benefit once we have a cache later (at least we'll be able to read cached K/Vs while we're waiting for etcd)

src/gravel/controllers/nodes/mgr.py
jecluis: I hope so. In this case was meant mostly as a PoC that watching keys works, but thinking ahead, given I hope we'll let the user refresh their token should they want to.

src/gravel/controllers/nodes/mgr.py
jecluis: Okay, I'm struggling to reproduce this with a script because it works fine with a Process. I truly can't tell wtf was going on. I'll try reverting it and see what happens, I guess.

jecluis:
@tserong addressed all comments, I think.

tserong (approved):

jhesketh (approved):
I haven't tested this manually, but I'm in favour of the general approach etc.

Issue #361, opened 2021-03-31
gravel: requires Python 3.7+ ? 


s0nea:
I'm not really sure if it's actually a bug and Python 3.7+ is required (and the issue can just be closed). Just in case it's not, I like to raise the issue here. So... on my local environment I noticed the following issue when running tox:

```
tox
py3 installed: aiofiles==0.6.0,attrs==20.3.0,click==7.1.2,dataclasses==0.8,fastapi==0.63.0,h11==0.12.0,importlib-metadata==3.10.0,iniconfig==1.1.1,packaging==20.9,pluggy==0.13.1,py==1.10.0,pydantic==1.7.3,pyfakefs==4.4.0,pyparsing==2.4.7,pytest==6.2.2,pytest-asyncio==0.14.0,pytest-datadir==1.3.1,pytest-mock==3.5.1,starlette==0.13.6,toml==0.10.2,typing-extensions==3.7.4.3,uvicorn==0.13.3,websockets==8.1,zipp==3.4.1
py3 runtests: PYTHONHASHSEED='2579776963'
py3 runtests: commands[0] | pytest gravel/
============================================================================================================ test session starts ============================================================================================================
platform linux -- Python 3.6.12, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
rootdir: /.../src/aquarium/src, configfile: tox.ini
plugins: datadir-1.3.1, pyfakefs-4.4.0, asyncio-0.14.0, mock-3.5.1
collected 17 items                                                                                                                                                                                                                          

gravel/tests/unit/cephadm/test_cephadm.py .......                                                                                                                                                                                     [ 41%]
gravel/tests/unit/controllers/test_config.py ...                                                                                                                                                                                      [ 58%]
gravel/tests/unit/controllers/test_gstate.py .F                                                                                                                                                                                       [ 70%]
gravel/tests/unit/controllers/orch/test_ceph.py .....                                                                                                                                                                                 [100%]

================================================================================================================= FAILURES ==================================================================================================================
_______________________________________________________________________________________________________________ test_tickers ________________________________________________________________________________________________________________

gstate = <gravel.controllers.gstate.GlobalState object at 0x7fdae58bfb00>

    @pytest.mark.asyncio
    async def test_tickers(gstate):
        from gravel.controllers.gstate import Ticker
    
        class TestTicker(Ticker):
            def __init__(self, name):
                super().__init__(name, 1.0)
                self.has_ticked = False
    
[...]

PR #405, opened 2021-04-12
tools: setup-dev.sh: check for python >= 3.8

tserong:
This ensures we've got at least python 3.8, which is what we use in our images, and is required for asyncio (it's not impossible 3.7 might work, but nobody's tested that so it seems safest to stick with 3.8).

I've also made it a bit more obvious when dependency installation fails.  I've seen cases where zypper fails to refresh some repo, but that scrolls off the top of the screen, the script bails out and all you see at the bottom is the "Nothing to do" line from zypper, so you just assume there's nothing to do.

## Checklist

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>

### DIFF
--- a/tools/setup-dev.sh
+++ b/tools/setup-dev.sh
@@ -129,7 +129,10 @@ if ! ${skip_install_deps} ; then
   case $osid in
     opensuse-tumbleweed | opensuse-leap)
       echo "=> try installing dependencies"
-      sudo zypper --non-interactive install ${dependencies_opensuse_tumbleweed[*]} || exit 1
+      sudo zypper --non-interactive install ${dependencies_opensuse_tumbleweed[*]} || {
+        echo "Dependency installation failed"
+        exit 1
+      }
       ;;
     debian)
       echo "=> installing nodejs15.x repo to apt source"
@@ -139,7 +142,10 @@ if ! ${skip_install_deps} ; then
       echo "=> installing kiwi repo to apt source"
       sudo add-apt-repository 'deb https://download.opensuse.org/repositories/Virtualization:/Appliances:/Builder/Debian_10 ./'
       echo "=> try installing dependencies"
-      sudo apt-get install -q -y ${dependencies_debian[*]} || exit 1
+      sudo apt-get install -q -y ${dependencies_debian[*]} || {
+        echo "Dependency installation failed"
+        exit 1
+      }
       ;;
     ubuntu)
       echo "=> installing nodejs15.x repo to apt source"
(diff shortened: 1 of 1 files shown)


tserong:
This breaks jenkins/setup-vagrant-leap, because leap only has python 3.6.12.

This commit was intended to ensure people hacking on Aquarium have the correct version of python. If you've got python 3.6, for example, the unit tests fail with "AttributeError: module 'asyncio' has no attribute 'create_task'" and the flake8 tests fail with "F407 future feature annotations is not defined".

Given we require python 3.8 for development, maybe we should stop testing on leap? What do y'all think?

jhesketh:
> This breaks jenkins/setup-vagrant-leap, because leap only has python 3.6.12.
> 
> This commit was intended to ensure people hacking on Aquarium have the correct version of python. If you've got python 3.6, for example, the unit tests fail with "AttributeError: module 'asyncio' has no attribute 'create_task'" and the flake8 tests fail with "F407 future feature annotations is not defined".
> 
> Given we require python 3.8 for development, maybe we should stop testing on leap? What do y'all think?

We've had a bit of this conversation on https://github.com/aquarist-labs/aquarium/issues/361

My opinion is that we should only support 3.8. We can provide some tips on how a developer might get that (eg through pyenv). If we expect a lot of developers to be using a particular path (eg pyenv on leap), then we should have our test validate that.

kshtsk:
Don't you think the setup-dev.sh script should have a --pyenv option in order to install pyenv from source and install python3.8, having it default on Leap?

kshtsk:
I wonder if we ask to use pyenv python3.8 how the librados system package can be used if it is 3.6 on Leap?

kshtsk:
```
> rpm -ql python3-rados
/usr/lib64/python3.6/site-packages/rados-2.0.0-py3.6.egg-info
/usr/lib64/python3.6/site-packages/rados-2.0.0-py3.6.egg-info/PKG-INFO
/usr/lib64/python3.6/site-packages/rados-2.0.0-py3.6.egg-info/SOURCES.txt
/usr/lib64/python3.6/site-packages/rados-2.0.0-py3.6.egg-info/dependency_links.txt
/usr/lib64/python3.6/site-packages/rados-2.0.0-py3.6.egg-info/top_level.txt
/usr/lib64/python3.6/site-packages/rados.cpython-36m-x86_64-linux-gnu.so
```

kshtsk:
and even if try and use pyenv that mean we should probably use librados from the ceph source tree, don't we?

jecluis (approved):
with #403 now merged, I believe this is mandatory.

tserong:
> with #403 now merged, I believe this is mandatory.

OK, merged. We can continue pyenv etc. discussion in #361

PR #417, opened 2021-04-13
tools/setup-dev.sh add pyenv support

kshtsk:
Add options to use pyenv instead of system python version.
Defaults to use pyenv on Leap, as well as install missing
python 3.8 version.

## Checklist

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>

### DIFF
--- a/tools/setup-dev.sh
+++ b/tools/setup-dev.sh
@@ -45,6 +45,21 @@ dependencies_ubuntu=(
   "libvirt-daemon"
 )
 
+dependencies_build_python_leap=(
+  "gcc"
+  "automake"
+  "bzip2"
+  "libbz2-devel"
+  "xz"
+  "xz-devel"
+  "openssl-devel"
+  "ncurses-devel"
+  "readline-devel"
+  "zlib-devel"
+  "libffi-devel"
+  "sqlite3-devel"
+)
+
 usage() {
   cat << EOF
 usage: $(basename $0) [options]
@@ -52,13 +67,15 @@ usage: $(basename $0) [options]
 options:
   --show-dependencies     Show required dependencies.
   --skip-install-deps     Skip installing dependencies.
+  --pyenv-python          Use pyenv python version, e.g. 3.8.9
+
   -h|--help               This message.
 EOF
 }
 
 yes_no() {
     while true; do
-        read -p "$1? " yn
+        read -p "$1 (y/n)? " yn
         case $yn in
             [Yy]* ) return 0; break;;
             [Nn]* ) return 1; break;;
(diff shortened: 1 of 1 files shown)


kshtsk:
jenkins run tumbleweed

kshtsk:
May resolve the issue #361 on Leap

AvengerMoJo:
look good, I need this finished PR and then I can add the rest of the distribution

tools/setup-dev.sh
```diff
@@ -171,7 +220,36 @@ if ! ${skip_install_deps} ; then
 
 fi
 
-if ! PY_VER_STR=$(python3 --version) ; then
+if $setup_pyenv ; then
```
AvengerMoJo, line 223: I don't think this is needed. Reduce the use and setup pyenv into a single var. If developer already has pyenv installed then just switch to the pyenv version to 3.8.9 default should work fine.

tserong (commented):
Works for me on Leap 15.1 (which I really must get around to upgrading to Leap 15.2...)

We should add these options to the help text (line 63+) and some information to doc/fzth, but see Alex's comment about maybe collapsing it to one option.

tools/setup-dev.sh
tserong: ...and if pyenv is not present, but use_pyenv is specified, then it could automatically go run the setup, right?

AvengerMoJo:
@kshtsk I did some update to your PR, maybe you can pull my branch and combine into a signle PR

tools/setup-dev.sh
AvengerMoJo: correct, I believe there will be only 3 cases. 
1) distribution already have default python3.8 packages 
2) no python3.8 default packages but developer already using pyenv 
3) no python3.8 default packages need us to install 

So with --use-pyenv flag we either default true to distribution we already know they don't have python3.8 or allow the user to pick the specific version

kshtsk: This option is needed, because are following cases:
1) --pyenv --setup-pyenv --python-pyenv 3.8.9 - First time 
2) --pyenv Sequential runs when we do not want even setup pyenv and rebuild python and we want to just rebuild venv
3) --pyenv --python-pyenv 3.8.8 if we do not want to drop existing .pyenv but want to switch and use to python 3.8.8

AvengerMoJo: if you look my patch it work the same way. 
--use-pyenv doesn't stop anyone from using even default package exist like TW. 
--use-pyenv doesn't reinstall if existed. 
--use-pyenv switch version base on --python-pyenv if exist. 
So we don't need to split the use and setup. Just use can cover all the cases.

kshtsk:
jenkins test leap

tserong (approved):
LGTM!

PR #418, opened 2021-04-14
images: config.sh: install aetcd3 direct from github

tserong:
Pip knows how to install Python packages directly from git repos, so let's take advantage of that for our tweaked aetcd3.

## Checklist

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>


tserong:
For reference: https://pip.pypa.io/en/stable/reference/pip_install/#git

tserong:
Dammit, hang on a sec...

jecluis:
It might be worth pulling directly from pypi. I read somewhere, subject to confirmation, that one can have personal repos for packages. That way we don't need a git dependency on the image. Unless folks find that useful for some other reason?

tserong:
Yeah, doing this _does_ mean adding git to the image, in order for the build to work, which might not be entirely desirable. My goal here is to make it a bit more obvious what the origin of the code is, and also to rely on public infrastructure. If we can do a personal repo on pypi, that might be better. Either way, I learned something, so, that's a win, I guess?

jecluis:
yep. Makes sense to me. We can add a pypi thingy later.

jecluis (approved):

PR #420, opened 2021-04-14
glass: Make sure all subscriptions are handled correctly

votdev:
- use 'take' RxJS operator to automatically unsubscribe the subscription.
- use RxJS 'timer' instead of setTimeout. This way it is easier to handle situations where the component is destroyed before the timer has been fired.
- unsubscribe subscriptions on component destruction. This way situations are handled where the subscription is done but has not been completed before the component is destroyed. Same applies to the timer subscription. It unsubscribes automatically after it has been fired once (thanks to 'take') and the component will unsubscribe if the component is destroyed before the timer has been finished.

## Checklist

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>

### DIFF
--- a/src/glass/src/app/shared/components/widget/widget.component.ts
+++ b/src/glass/src/app/shared/components/widget/widget.component.ts
@@ -1,7 +1,7 @@
 import { Component, EventEmitter, Input, OnDestroy, OnInit, Output } from '@angular/core';
 import * as _ from 'lodash';
-import { EMPTY, Observable, Subscription } from 'rxjs';
-import { catchError, finalize } from 'rxjs/operators';
+import { EMPTY, Observable, Subscription, timer } from 'rxjs';
+import { catchError, finalize, take } from 'rxjs/operators';
 
 export type WidgetAction = {
   icon: string;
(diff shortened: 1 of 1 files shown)


src/glass/src/app/shared/components/widget/widget.component.ts
```diff
@@ -62,14 +64,14 @@ export class WidgetComponent implements OnInit, OnDestroy {
         finalize(() => {
           this.error = !this.loadingWithoutError;
           this.firstLoadComplete = this.loadingWithoutError;
-          this.refreshDataSubscription?.unsubscribe();
           if (this.reloadTime > 0) {
-            setTimeout(() => {
-              this.reload();
-            }, this.reloadTime);
+            this.timerSubscription = timer(this.reloadTime)
+              .pipe(take(1))
```
Devp00l, line 69: Maybe it's nicer to use [first()](https://rxjs-dev.firebaseapp.com/api/operators/first) here instead of `take(1)`.

votdev: I wanted to use the `first` operator at first, but then i thought `take` expresses more the intention of this operator. What do you think?

Devp00l: I thought `first` would be easier to understand, but you think the opposite. To not get into a bike shed discussion than - just leave it as it is ;)

Devp00l (approved):

PR #431, opened 2021-04-16
tools/setup-dev.sh: remove python3-rados dependency

tserong:
python3-rados is only be needed at runtime, in the image. It shouldn't be necessary for this to be installed on the host, and it makes life easier if it's not, when using different python versions via pyenv.

## Checklist

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>

### DIFF
--- a/tools/setup-dev.sh
+++ b/tools/setup-dev.sh
@@ -6,7 +6,6 @@ dependencies_opensuse_tumbleweed=(
   "kpartx"
   "make"
   "python3"
-  "python3-rados"
   "python3-kiwi"
   "nodejs-common"
   "npm"
@@ -21,7 +20,6 @@ dependencies_debian=(
   "make"
   "python3"
   "python3-pip"
-  "python3-rados"
   "python3-kiwi"
   "python3-kiwi-boxed-plugin"
   "python3-venv"
@@ -36,7 +34,6 @@ dependencies_ubuntu=(
   "make"
   "python3"
   "python3-pip"
-  "python3-rados"
   "python3-kiwi"
   "python3-kiwi-boxed-plugin"
   "python3-venv"
@@ -291,11 +288,6 @@ if [ $PY_MINOR -lt 8 ] ; then
   exit 1
 fi
 
-if ! ( echo "import rados" | python3 &>/dev/null ); then
-  echo "error: missing rados python bindings"
-  exit 1
-fi
-
 if ! npm --version &>/dev/null ; then
   echo "error: missing npm"
   exit 1
@@ -318,9 +310,11 @@ if [ -d venv ] ; then
 fi
 
 
-# we need system site packages because librados python bindings appear to only
-# be available as a package. It might be a good idea to compile it from the
-# ceph repo we keep as a submodule, but it might be overkill at the moment?
+# We need system site packages because librados python bindings are only
+# available as a package, they're not on pypi (and Tim thinks they probably
+# never will be). Setting --system-site-packages here means the venv will
+# work properly if someone uses it inside the image, where we need to be
+# able to fall back to system packages to get librados.
 $PYTHON -m venv --system-site-packages venv || exit 1
 
 source venv/bin/activate


tserong:
OK, that's odd. @kshtsk is it possible to find out why `systemctl start libvirtd` failed in https://ceph-ci.suse.de/job/aquarium-setup-vagrant-tumbleweed/227/consoleFull? Or should I maybe try adding something like the following (untested) code?

```diff
--- a/.wasser/config.yaml
+++ b/.wasser/config.yaml
@@ -52,7 +52,11 @@ routines:
         END
     - name: make sure libvirtd is running
       command: |
-        sudo systemctl start libvirtd
+        ( sudo systemctl start libvirtd || {
+          systemctl status libvirtd.service
+          journalctl -xe
+          exit 1
+        } )
     - name: run microos image on vagrant
       command: |
         cd aquarium
```

tools/setup-dev.sh
```diff
@@ -318,9 +310,11 @@ if [ -d venv ] ; then
...
 
-# we need system site packages because librados python bindings appear to only
-# be available as a package. It might be a good idea to compile it from the
-# ceph repo we keep as a submodule, but it might be overkill at the moment?
+# We need system site packages because librados python bindings are only
+# available as a package, they're not on pypi (and Tim thinks they probably
+# never will be). Setting --system-site-packages here means the venv will
+# work properly if someone uses it inside the image, where we need to be
+# able to fall back to system packages to get librados.
 $PYTHON -m venv --system-site-packages venv || exit 1
```
kshtsk, line 318: I really don't understand how host's venv can help on the guest VM, why do I need this.

kshtsk:
jenkins start tumbleweed

kshtsk:
> OK, that's odd. @kshtsk is it possible to find out why `systemctl start libvirtd` failed in https://ceph-ci.suse.de/job/aquarium-setup-vagrant-tumbleweed/227/consoleFull? Or should I maybe try adding something like the following (untested) code?
> 
> ```diff
> --- a/.wasser/config.yaml
> +++ b/.wasser/config.yaml
> @@ -52,7 +52,11 @@ routines:
>          END
>      - name: make sure libvirtd is running
>        command: |
> -        sudo systemctl start libvirtd
> +        ( sudo systemctl start libvirtd || {
> +          systemctl status libvirtd.service
> +          journalctl -xe
> +          exit 1
> +        } )
>      - name: run microos image on vagrant
>        command: |
>          cd aquarium
> ```

I guess we just can kick it login

kshtsk:
It is reproducible with other PRs, something bad happened with Tumbleweed, needs investigation.

kshtsk:
```Apr 16 10:23:07 pr-431-tumbleweed systemd[1]: Starting Virtualization daemon...
Apr 16 10:23:07 pr-431-tumbleweed libvirtd[19268]: /usr/sbin/libvirtd: symbol lookup error: /usr/lib64/libgobject-2.0.so.0: undefined symbol: g_memdup2
Apr 16 10:23:07 pr-431-tumbleweed systemd[1]: libvirtd.service: Main process exited, code=exited, status=127/n/a
Apr 16 10:23:07 pr-431-tumbleweed systemd[1]: libvirtd.service: Failed with result 'exit-code'.
Apr 16 10:23:07 pr-431-tumbleweed systemd[1]: Failed to start Virtualization daemon.
Apr 16 10:23:07 pr-431-tumbleweed systemd[1]: libvirtd.service: Scheduled restart job, restart counter is at 5.
Apr 16 10:23:07 pr-431-tumbleweed systemd[1]: Stopped Virtualization daemon.
Apr 16 10:23:07 pr-431-tumbleweed systemd[1]: libvirtd.service: Start request repeated too quickly.
Apr 16 10:23:07 pr-431-tumbleweed systemd[1]: libvirtd.service: Failed with result 'exit-code'.
Apr 16 10:23:07 pr-431-tumbleweed systemd[1]: Failed to start Virtualization daemon.
```

kshtsk:
that is what when I am saying tumbleweed is not stable.

kshtsk:
I guess we need just fresher image of tumbleweed or us `zypper dup`.

kshtsk:
or both

kshtsk:
jenkins run tumbleweed

tools/setup-dev.sh
tserong: This is a bit obscure... If we create the venv with `--system-site-packages`, it sets `include-system-site-packages = true` inside `venv/pyvenv.cfg`, so, later, if you _use_ the venv from inside the image, this means it will correctly pick up python3-rados from the system packages inside the image. If we created the venv without that option, using the venv inside the image wouldn't work.

That said, IMO you shouldn't usually need to use the venv inside the image at all (because everything is installed system wide already), so it's a bit of a moot point, but, when chatting to @votdev the other day, he said he _does_ use the venv inside the image sometimes, to handle added dependencies easily, without having to rebuild the image.

tserong:
> that is what when I am saying tumbleweed is not stable.

Dammit. Thanks for investigating.

kshtsk:
jenkins run tumbleweed

kshtsk:
jenkins run tumbleweed

tools/setup-dev.sh
kshtsk: I see your point, but in general we MUST NOT use the venv backed on the host inside virtual images. So if one wants to have venv inside the image it should be created right in the guest system in order to avoid any kind of misfit of packages and modules.

tserong: Right, makes sense.

kshtsk:
@tserong I've switched to the latest Tumbleweed image and it passed.

tserong:
> @tserong I've switched to the latest Tumbleweed image and it passed.

Awesome, thanks.

tools/setup-dev.sh
tserong: If we want to force setup of a proper inside-image venv, then I think we should do that in run_aquarium.sh, which we can do in a separate PR.

kshtsk (approved):

PR #435, opened 2021-04-16
requirements: use aquarist-labs's aetcd3 fork

jecluis:
And fix to a specific sha instead of a branch name.

### DIFF
--- a/images/microOS/config.sh
+++ b/images/microOS/config.sh
@@ -175,7 +175,7 @@ fi
 
 if [[ "$kiwi_profiles" == *"Ceph"* ]]; then
   pip install fastapi uvicorn \
-    git+https://github.com/jecluis/aetcd3/@wip-aquarium#egg=aetcd3
+    git+https://github.com/aquarist-labs/aetcd3/@7b68cdd9c8b436b0b0a49dab9984174670ca1e30
   baseInsertService aquarium
 fi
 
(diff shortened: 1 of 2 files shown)


kshtsk:
jenkins run tumbleweed

jhesketh (approved):
Would it be better to pin to a tag? Either way lgtm.

jecluis:
I'll be happy to pin it to a tag if we end up carrying this fork and depending on it. Until then, I'd rather not introducing our tags on that history. Maybe I'm being overzealous/paranoid :shrug:

tserong (commented):
This removes `#egg=aetcd3` from the end of the URLs. Does that matter? (I assume it doesn't matter, or the tests would have failed? But https://pip.pypa.io/en/stable/reference/pip_install/#git always puts #egg=... on the end so thought I'd check).

I'm ambivalent about whether we use a SHA or tag for this, but agree that either one is preferable to a branch (the head of which might change).

jecluis:
> This removes `#egg=aetcd3` from the end of the URLs. Does that matter? (I assume it doesn't matter, or the tests would have failed? But https://pip.pypa.io/en/stable/reference/pip_install/#git always puts #egg=... on the end so thought I'd check).

It does. pip was throwing a fit keeping that there because I had this brilliant idea of changing the package's name to `aquarium-aetcd3` in `setup.py` and the egg name and the package's name did not match. So I dropped the one that was less annoying.

Now the library is installed in an `aquarium-etcd` directory, but the library is still used using the `aetcd3` package.

PR #436, opened 2021-04-17
gravel: make node join work properly with etcd, drop node roles

### LINKED ISSUE #268: Remove node roles (i.e. no leaders, no followers, everything is just an Aquarium node)
This is resulting from discussion in https://github.com/aquarist-labs/aquarium/pull/266#issuecomment-793513870. Right now (2021-03-09) the first node bootstrapped is the Leader and other nodes are Followers, and need to join the cluster by talking to that Leader. I'd like to ideally remove this concept, so that any new node can talk to any existing node to join the cluster, i.e. there's no concept of a leader or followers, except perhaps for the duration of a join (where the joining node may need to have a brief discussion with whichever node it's talking to in order to actually do the join), but with that exception, every node is equal, and effectively stateless. If we need to know which nodes are present, we can ask Ceph. If Ceph is down, well, we can't add new nodes anyway, so the list of existing nodes becomes somewhat moot at that point ;-)

Note that this issue relates to https://github.com/aquarist-labs/aquarium/issues/32 - just because I'd like everything to be stateless, doesn't mean we don't somehow need to share state...

jecluis:
We had a tiny problem when expanding the cluster: we were providing wrong etcd initial peers to the joining node; instead of using the information provided by aetcd3's `client.add_member()`, we were assuming what those values could be based on our limited knowledge of the cluster. Even though that information was enough to add a second node, it was not enough to add three or more nodes.

Instead, we now rely on additional return from `client.add_member()`, which provides us with a list of existing cluster members, which we then transform into something the joining node can use. The only snafu is that aetcd3, at this point in time, does not support this additional information being returned by said function: we had to patch aetcd3 (and [opened a PR upstream](https://github.com/martyanov/aetcd3/pull/7)) to obtain this additional information.

This means we rely on a custom version of aetcd3, living in our organization's aetcd3 fork. Because we are already using a custom version, this PR should not be merged before #435 or we will risk breaking things for folks multiple times. This PR references a SHA that, while it lives in a branch in our fork, is not yet committed to the `aquarium` branch. Once #435 is merged, we can then merge https://github.com/aquarist-labs/aetcd3/pull/1 , and then adjust this PRs SHA and finally merge.

Additionally, we finally decided to make @tserong happy and dropped node-specific roles. No longer do we have a leader and a follower. A new node can join the cluster by calling on any existing node. This needs to be enhanced so a node does not allow a join if it's not in quorum (e.g., etcd is down, or its etcd is unable to reach the remaining of the cluster for whatever reason), but that's something to be dealt with once we start fiddling with etcd-driven node state.

Resolves #268 

### DIFF
--- a/images/microOS/config.sh
+++ b/images/microOS/config.sh
@@ -175,7 +175,7 @@ fi
 
 if [[ "$kiwi_profiles" == *"Ceph"* ]]; then
   pip install fastapi uvicorn \
-    git+https://github.com/aquarist-labs/aetcd3/@7b68cdd9c8b436b0b0a49dab9984174670ca1e30
+    git+https://github.com/aquarist-labs/aetcd3/@edf633045ce61c7bbac4d4a6ca15b14f8acfe9cd
   baseInsertService aquarium
 fi
 
--- a/src/requirements.txt
+++ b/src/requirements.txt
@@ -7,4 +7,4 @@ starlette==0.13.6
 uvicorn==0.13.3
 pip
 websockets==8.1
-git+https://github.com/aquarist-labs/aetcd3/@7b68cdd9c8b436b0b0a49dab9984174670ca1e30
+git+https://github.com/aquarist-labs/aetcd3/@edf633045ce61c7bbac4d4a6ca15b14f8acfe9cd
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -1,11 +1,11 @@
-name: tox 
+name: "gravel tests" 
 
 on:
   - push
   - pull_request
 
 jobs:
-  build:
+  tox:
     runs-on: ubuntu-latest
     strategy:
       matrix:
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3]
+        python-version: [3.8]
 
     steps:
     - uses: actions/checkout@v2
--- a/src/gravel/controllers/nodes/mgr.py
+++ b/src/gravel/controllers/nodes/mgr.py
@@ -74,12 +74,6 @@ from gravel.controllers.orch.orchestrator import Orchestrator
 logger: Logger = fastapi_logger
 
 
-class NodeRoleEnum(int, Enum):
-    NONE = 0
-    LEADER = 1
-    FOLLOWER = 2
-
-
 class NodeStageEnum(int, Enum):
     NONE = 0
     BOOTSTRAPPING = 1
@@ -99,7 +93,6 @@ class NodeInitStage(int, Enum):
 
 class NodeStateModel(BaseModel):
     uuid: UUID
-    role: NodeRoleEnum
     stage: NodeStageEnum
     address: Optional[str]
     hostname: Optional[str]
@@ -196,7 +189,6 @@ class NodeMgr:
 
             state = NodeStateModel(
                 uuid=uuid4(),
-                role=NodeRoleEnum.NONE,
                 stage=NodeStageEnum.NONE,
                 address=None,
                 hostname=None
@@ -240,7 +232,6 @@ class NodeMgr:
         assert self._state
         assert self._state.stage == NodeStageEnum.READY or \
             self._state.stage == NodeStageEnum.BOOTSTRAPPED
-        assert self._state.role != NodeRoleEnum.NONE
 
         logger.info("start node")
 
@@ -248,8 +239,6 @@ class NodeMgr:
         await self._load()
 
         self._init_stage = NodeInitStage.STARTED
-        if self._state.role != NodeRoleEnum.LEADER:
-            return
 
         logger.info("start leader node")
         self._incoming_task = asyncio.create_task(self._incoming_msg_task())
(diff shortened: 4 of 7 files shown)


jecluis:
@kshtsk latest commit changes tox's version to python 3.8, as well as the github workflow for tox. Mind checking that out?

kshtsk:
jenkins run tumbleweed

kshtsk:
@jecluis that looks great for me, however did you think something specific to check?

jecluis:
@kshtsk just wanted to make sure you'd be aware of the changes to the github and tox stuff, given it's somewhat unrelated to this PR (except for the part that the PR is broken without it).

kshtsk:
Hm, btw, can maybe rename "build" job to some thing appropriate? Like 'python' or 'python_unittest'?
Btw, the 'build(3)' is currently 'required', and I guess we need to change 'main' rule to use 'build(3.8)' or whatever we rename.

jecluis:
I'll be happy to. Mind pointing me to where the build job is described/configured? Doesn't seem to be part of the workflows kept in `.github/`.

jecluis:
Nevermind, I think I got it. Should be in the `tox` workflow.

jecluis:
@kshtsk this one is the one I really can't figure out where to find:

image

kshtsk:
> @kshtsk this one is the one I really can't figure out where to find:
> 
> image

This is what I am talking about.
When you go to the "aquarium" project Settings -> Branches -> Branch protection rules -> Click "Edit" for "main" -> Require status checks to pass before merging -> Observe checkbox for "build (3)", which checked and marked "Required". Once you uncheck it, will not be required anymore for all PRs.

jecluis:
> > @kshtsk this one is the one I really can't figure out where to find:
> > image
> 
> This is what I am talking about.
> When you go to the "aquarium" project Settings -> Branches -> Branch protection rules -> Click "Edit" for "main" -> Require status checks to pass before merging -> Observe checkbox for "build (3)", which checked and marked "Required". Once you uncheck it, will not be required anymore for all PRs.

Alright, got it. But I still don't know where the "build" check comes from. I understand from where pretty much every other check on that list comes from, but the "build (3)" and "build (3.8)" come from is a mystery to me. Any idea?

kshtsk:
> > > @kshtsk this one is the one I really can't figure out where to find:
> > > image
> > 
> > 
> > This is what I am talking about.
> > When you go to the "aquarium" project Settings -> Branches -> Branch protection rules -> Click "Edit" for "main" -> Require status checks to pass before merging -> Observe checkbox for "build (3)", which checked and marked "Required". Once you uncheck it, will not be required anymore for all PRs.
> 
> Alright, got it. But I still don't know where the "build" check comes from. I understand from where pretty much every other check on that list comes from, but where the "build (3)" and "build (3.8)" come from is a mystery to me. Any idea?

It has been comming from the line: https://github.com/aquarist-labs/aquarium/pull/436/files#diff-8a1944567ec5d9e12aaaac28489c47e6567e13b58d861815379f9b8d3aeb2623L8

Since the python version from the matrix was 3, the check had the title "build (3)". This change is turning it to "tox (3.8)".

jecluis:
Okay, so, to merge this PR it needs to be disabled; and, once this PR is merged, that check should disappear from main?

kshtsk:
> Okay, so, to merge this PR it needs to be disabled; and, once this PR is merged, that check should disappear from main?

Yes. But we can add the new tox name and by the way, glass unit tests nodejs tests we can add as required.

images/microOS/config.sh
```diff
@@ -175,7 +175,7 @@ fi
 
 if [[ "$kiwi_profiles" == *"Ceph"* ]]; then
   pip install fastapi uvicorn \
-    git+https://github.com/jecluis/aetcd3/@wip-aquarium#egg=aetcd3
```
kshtsk, line 178: nit: I think using tags is more suitable than weird sha1.

src/requirements.txt
```diff
@@ -7,4 +7,4 @@ starlette==0.13.6
 uvicorn==0.13.3
 pip
 websockets==8.1
-git+https://github.com/jecluis/aetcd3/@wip-aquarium#egg=aetcd3
```
kshtsk, line 10: nit: same here, using a tag is less cryptic, and it can point to any sha1.

kshtsk (dismissed):

kshtsk:
@jecluis I have removed the tox build job from required for now, and you can see it is disappeared from the blockers for this PR. However the other PRs do not require it any longer.

images/microOS/config.sh
jecluis: I agree, but I don't think we should be tagging the repo if we keep consuming from aetcd3's upstream, or we could end up with conflicting tags, or just making the whole history weird. Also, tagging just because one wants to merge a PR seems weirdly silly.

tserong (dismissed):
Needs conflicts resolved since #435 was merged, but otherwise LGTM

tserong (approved):

jecluis:
tumbleweed test is stuck.

jecluis:
jenkins run tumbleweed

jecluis:
I can't tell whether tumbleweed is stuck again or not, but I'm fairly certain everything is okay anyway (leap passed, everything was okay before the rebase, trivial rebase). I'm going to merge anyway, and deal with any fall-out should that happen. Forgive me @kshtsk

kshtsk:
@kshtsk that's fine, I don't know what is problem, but some times python Thread for one of the stream got stuck for the git clone command.

Issue #426, opened 2021-04-14
glass: CephfsModalComponent and NfsModalComponent refactoring


s0nea:
The `CephfsModalComponent` and `NfsModalComponent` are quite similar. Add a more general component in order to get rid of the duplicated code.

Combine both components into a `FileServiceModalComponent` and add a selection box containing `CephFS` and `NFS` options.

jecluis:
Can we increase the prio on this one from a Nice To Have to a Required? @votdev ? :)

PR #448, opened 2021-04-22
glass: refactor `NFS`/`CephFS` components

s0nea:
The components providing the `NFS` and `CephFS` services are pretty similar. Refactor both in order to combine them
into one `File` component.

Fixes: https://github.com/aquarist-labs/aquarium/issues/426

## Checklist

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>


src/glass/src/app/pages/deployment-page/deployment-page.component.html
```diff
@@ -106,7 +106,7 @@
           {{ 'Configure CephFS service' | translate }}
           <mat-icon class="glass-color-accent"
                     svgIcon="mdi:plus-circle-outline"
-                    (click)="openCephfsDialog()">
+                    (click)="openFileServiceDialog('cephfs')">
```
jecluis, line 109: Something to keep in mind for the future, it would be nice if we could get rid of adding NFS and CephFS services individually and just had the possibility of adding a "FileService"; especially given the possibility of choosing the backing store for the service, it feels a bit moot still supporting adding services individually.

jecluis (approved):
LGTM, although I haven't tested it yet.

I feel this could be a good opportunity to remove the possibility of adding individual nfs and cephfs services though, and instead only having, adding, and showing a File service which would have either an 'nfs' or a 'cephfs' backing store. But I also think that could be done in a follow up PR.

votdev (approved):

PR #449, opened 2021-04-22
gravel: run etcd within a container

mgfritch:
* reduces the number of system (e.g. rpm) dependencies for the image
* consistent with platforms such as CoreOS and k8s
* could potentially make the orchestration of etcd a bit easier later down the road ...

## Checklist

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>

### DIFF
--- a/src/gravel/controllers/config.py
+++ b/src/gravel/controllers/config.py
@@ -48,6 +48,12 @@ class ServicesOptionsModel(BaseModel):
     probe_interval: float = Field(1.0, title="Services Probe Interval")
 
 
+class EtcdOptionsModel(BaseModel):
+    registry: str = Field("quay.io/coreos/etcd", title="Container Image Registry")
+    version: str = Field("latest", title="Container Version Label")
+    data_dir: str = Field("/var/lib/etcd", title="Etcd Data Dir")
+
+
 class OptionsModel(BaseModel):
     service_state_path: Path = Field(Path(config_dir).joinpath("storage.json"),
                                      title="Path to Service State file")
@@ -56,6 +62,7 @@ class OptionsModel(BaseModel):
     devices: DevicesOptionsModel = Field(DevicesOptionsModel())
     status: StatusOptionsModel = Field(StatusOptionsModel())
     services: ServicesOptionsModel = Field(ServicesOptionsModel())
+    etcd: EtcdOptionsModel = Field(EtcdOptionsModel())
 
 
 class ConfigModel(BaseModel):
--- a/src/gravel/controllers/nodes/mgr.py
+++ b/src/gravel/controllers/nodes/mgr.py
@@ -412,6 +412,10 @@ class NodeMgr:
         hostname = self._state.hostname
         address = self._state.address
 
+        data_dir: Path = Path("/var/lib/etcd")
+        if not data_dir.exists():
+            data_dir.mkdir(0o700)
+
         logger.info(f"starting etcd, hostname: {hostname}, addr: {address}")
 
         def _get_etcd_args() -> List[str]:
@@ -429,7 +433,7 @@ class NodeMgr:
                 "--listen-client-urls", f"{client_url},http://127.0.0.1:2379",
                 "--advertise-client-urls", client_url,
                 "--initial-cluster", initial_cluster,
-                "--data-dir", f"/var/lib/etcd/{hostname}.etcd"
+                "--data-dir", str(data_dir),
             ]
 
             if new:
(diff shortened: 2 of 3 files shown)


tserong (commented):
LGTM, but can you please remind me why we're containerizing etcd? I'm sure we discussed this on the call the other day, but it's somehow completely evaporated from my brain.

Also, I assume this means we can remove the etcd (and etcdctl?) packages from images/microOS/config.xml

jecluis:
> LGTM, but can you please remind me why we're containerizing etcd? I'm sure we discussed this on the call the other day, but it's somehow completely evaporated from my brain.

One of the reasons that came up during the call was resource limitation for etcd, although I'm not entirely sure that's something we would want to do. Other than that, I don't particularly see a benefit other than saying we're containerizing as much as we can.

> Also, I assume this means we can remove the etcd (and etcdctl?) packages from images/microOS/config.xml

Yeah, which one the one hand reduces our hard dependencies, which is cool, but on the other is making Aquarium dependent on outside resources to work. :shrug:

src/gravel/controllers/nodes/mgr.py
```diff
@@ -412,39 +412,61 @@ async def _spawn_etcd(
...
+
+            if new:
+                assert token
+                args += ["--initial-cluster-state", "new"]
+                args += ["--initial-cluster-token", token]
+
+            return args
+
+        def _get_container_cmd():
+            registry = "quay.io/coreos/etcd"
```
jecluis, line 448: this should be a config option imo

jecluis (approved):
The commit should have a description of why this is preferable over spawning the etcd process, but overall looks good to me.

src/gravel/controllers/nodes/mgr.py
mgfritch: etcd conf options added for `registry`, `version`, and `data_dir`.

mgfritch:
> Also, I assume this means we can remove the etcd (and etcdctl?) packages from images/microOS/config.xml

@tserong  thanks, I completely overlooked that.


`etcd` is now removed, but I left `etcdctl` in the image. We could remove `etcdctl` too, but that would require us to use `etcdctl` from a container image which is likely less convenient for debugging ... ?

src/gravel/controllers/config.py
```diff
@@ -56,6 +62,7 @@ class OptionsModel(BaseModel):
     devices: DevicesOptionsModel = Field(DevicesOptionsModel())
     status: StatusOptionsModel = Field(StatusOptionsModel())
     services: ServicesOptionsModel = Field(ServicesOptionsModel())
+    etcd: EtcdOptionsModel = Field(EtcdOptionsModel())
```
mgfritch, line 65: perhaps should be a standalone `etcd.json` file??

kshtsk:
jenkins start tumbleweed

src/gravel/controllers/nodes/mgr.py
```diff
@@ -414,35 +414,35 @@ async def _spawn_etcd(
...
+                "--initial-cluster", initial_cluster,
+                "--initial-cluster-state", "existing",
+                "--data-dir", f"/var/lib/etcd/{hostname}.etcd"
+            ]
 
             if new:
                 assert token
-                args_dict["initial-cluster-token"] = token
-                args_dict["initial-cluster-state"] = "new"
+                args += ["--initial-cluster-state", "new"]
```
tserong, line 442: I think this means args will contain both `--initial-cluster-state existing` _and_ `--initial-cluster-state new` in the new case. Is that going to be a problem?

src/gravel/controllers/config.py
tserong: Dunno... It's etcd as spawned by Aquarium, so we own it, so it's OK to keep it inside the larger Aquarium config? (I don't really have an opinion)

tserong:
> `etcd` is now removed, but I left `etcdctl` in the image. We could remove `etcdctl` too, but that would require us to use `etcdctl` from a container image which is likely less convenient for debugging ... ?

Agreed.

kshtsk:
btw, tox has failure for flake and mypy

kshtsk:
jenkins test tumbleweed

src/gravel/controllers/config.py
jecluis: I'd leave it in our own config until there's a strong reason not to.

src/gravel/controllers/nodes/mgr.py
jecluis: Right, this might or might not work depending on how etcd deals with overriding option values. I'd rather do this right on our end instead of hoping etcd does what we think should happen.

jecluis (changes requested):
Generally looks good to me, except for a couple of comments.

Additionally, would be nice if the commit moving to a container had a brief description of why it's cool to have it in a container; not just the PR's description. Would not hold because of that.

Finally, would be even cooler if this particular reason was expanded a bit more:

> could potentially make the orchestration of etcd a bit easier later down the road ...

Especially because I, for one, don't see a particular benefit given how we're using etcd. So it would be nice to fully understand this point of view.

mgfritch:
> btw, tox has failure for flake and mypy

fixed!

src/gravel/controllers/nodes/mgr.py
mgfritch: oops, good catch! hanks @tserong

jecluis (approved):

PR #459, opened 2021-04-28
gravel: rework, cleanup node deployment

jecluis:
This is just a first pass towards a proper rework of the initial node deployment workflow.

Right now we're simply spinning-off a bit of the bootstrap and join code to a `deployment.py` module.

I'm going to do this in batches to make it easier to review.

### DIFF
--- a/src/gravel/api/bootstrap.py
+++ b/src/gravel/api/bootstrap.py
@@ -89,6 +89,7 @@ async def start_bootstrap(request: Request) -> StartReplyModel:
             message=e.message
         )
     except Exception as e:
+        logger.exception(e)
         logger.error(f"[API] unknown error on bootstrap: {str(e)}")
         success = False
         error = BootstrapErrorModel(
--- a/src/gravel/controllers/errors.py
+++ b/src/gravel/controllers/errors.py
@@ -1,0 +1,24 @@
+# project aquarium's backend
+# Copyright (C) 2021 SUSE, LLC.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+from typing import Optional
+
+
+class GravelError(Exception):
+    def __init__(self, msg: Optional[str] = ""):
+        super().__init__()
+        self._msg = msg
+
+    @property
+    def message(self) -> str:
+        return self._msg if self._msg else "gravel error"
--- a/src/gravel/controllers/nodes/deployment.py
+++ b/src/gravel/controllers/nodes/deployment.py
@@ -1,0 +1,126 @@
+# project aquarium's backend
+# Copyright (C) 2021 SUSE, LLC.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+
+from enum import Enum
+from pydantic import BaseModel, Field
+
+from gravel.controllers.errors import GravelError
+from gravel.controllers.gstate import GlobalState
+
+
+class NodeStageEnum(int, Enum):
+    NONE = 0
+    BOOTSTRAPPING = 1
+    DEPLOYED = 2
+    JOINING = 3
+    READY = 4
+    ERROR = 5
+
+
+class DeploymentModel(BaseModel):
+    stage: NodeStageEnum = Field(NodeStageEnum.NONE)
+
+
+class DeploymentError(GravelError):
+    pass
+
+
+class DeploymentState:
+
+    _stage: NodeStageEnum
+    _gstate: GlobalState
+
+    def __init__(self, gstate: GlobalState):
+        self._gstate = gstate
+        self._stage = NodeStageEnum.NONE
+
+        self._load_stage()
+
+        pass
+
+    def _load_stage(self) -> None:
+        try:
+            dep = self._gstate.config.read_model("deployment", DeploymentModel)
+            self._stage = dep.stage
+        except FileNotFoundError:
+            self._gstate.config.write_model("deployment", DeploymentModel())
+
+    def _save_stage(self) -> None:
+        try:
+            self._gstate.config.write_model(
+                "deployment",
+                DeploymentModel(stage=self._stage)
+            )
+        except Exception as e:
+            raise DeploymentError(str(e))
+
+    @property
+    def stage(self) -> NodeStageEnum:
+        return self._stage
+
+    @property
+    def nostage(self) -> bool:
+        return self._stage == NodeStageEnum.NONE
+
+    @property
+    def bootstrapping(self) -> bool:
+        return self._stage == NodeStageEnum.BOOTSTRAPPING
+
+    @property
+    def joining(self) -> bool:
+        return self._stage == NodeStageEnum.JOINING
+
+    @property
+    def deployed(self) -> bool:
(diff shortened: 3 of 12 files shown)


src/gravel/api/bootstrap.py
```diff
@@ -89,6 +89,7 @@ async def start_bootstrap(request: Request) -> StartReplyModel:
             message=e.message
         )
     except Exception as e:
+        logger.exception(e)
         logger.error(f"[API] unknown error on bootstrap: {str(e)}")
```
jhesketh, line 93: (nit) I believe you can just replace the above two lines with:
`logger.exception("[API] unknown error on bootstrap")`
(The exception and original message will also be logged)

src/gravel/controllers/errors.py
```diff
@@ -0,0 +1,24 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+from typing import Optional
+
+
+class GravelError(Exception):
```
jhesketh, line 17: (nit) if this is a "gravel" error, should it be in `src/gravel/errors.py`, rather than limiting it to controllers?

src/gravel/controllers/nodes/deployment.py
```diff
@@ -265,3 +288,79 @@ async def join(
...
+
+        try:
+            await self._bootstrapper.bootstrap(
+                address, finish_bootstrap_cb
+            )
+        except BootstrapError as e:
+            logger.error(f"bootstrap error: {e.message}")
+            raise NodeCantBootstrapError(e.message)
+
+    async def _prepare_bootstrap(
```
jhesketh, line 338: (nit since this is carried over) I'd really like to get rid of "prepare bootstrap". Bootstrapping is preparing, so this is like "bootstrap bootstrap" and so it bothers me (totally a nit). It could just as well be "prepare_etcd"?

src/gravel/controllers/nodes/deployment.py
```diff
@@ -265,3 +288,79 @@ async def join(
...
+
+        await spawn_etcd(
+            self._gstate,
+            new=True,
+            token=token,
+            hostname=hostname,
+            address=address
+        )
+
+    async def _start_bootstrap(self) -> None:
```
jhesketh, line 358: (nit) Similarly I find this confusing. I know it's denoted as a private method, but if you ignore that it's confusing if I should "bootstrap" the node or "start the bootstrap".. This could probably just be folded into the bootstrap method rather than being its own method.

src/gravel/controllers/nodes/deployment.py
```diff
@@ -265,3 +288,79 @@ async def join(
...
+        assert self._state
+        assert self._state.nostage
+        self._state.mark_bootstrap()
+
+    def finish_bootstrap(self) -> None:
```
jhesketh, line 363: (nit) ditto previous comment

jhesketh (approved):
lgtm, just a few nits that could be done as a followup or ignored

src/gravel/controllers/nodes/deployment.py
jecluis: I get that, but _bootstrap_ is also an action. You can prepare to start said action, start said action, and then finish said action. Aside from putting all the functions in the bootstrap function's context, I don't know a decent way out of this.

jhesketh: why not just call it prepare_etcd?

(again, this is a nit, so I'm not really fussed)

jecluis: works for this function, sure; but I was addressing the other comments as well ;)

jhesketh: Ah yes, in those cases I had just suggested rolling them into the one bootstrap.

src/gravel/controllers/nodes/deployment.py
jecluis: this should actually be a `finish_deployment()` instead.

src/gravel/controllers/nodes/deployment.py
jecluis: I guess they can be moved under `bootstrap()`'s context :shrug:

src/gravel/controllers/utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+from pathlib import Path
+from typing import Type, TypeVar
+from pydantic import BaseModel
+from pydantic.tools import parse_file_as
+
+
+def _get_file_path(dirpath: Path, name: str) -> Path:
+    if not dirpath.exists() or not dirpath.is_dir():
+        raise NotADirectoryError()
```
mgfritch, line 23: Unhandled exception? should we attempt to create the dirpath if it does not exist?

src/gravel/controllers/nodes/deployment.py
```diff
@@ -0,0 +1,126 @@
...
+        return (
+            self._stage == NodeStageEnum.NONE or
+            self._stage == NodeStageEnum.DEPLOYED or
+            self._stage == NodeStageEnum.READY
+        )
+
+    def mark_bootstrap(self) -> None:
+        assert self.nostage
+        self._stage = NodeStageEnum.BOOTSTRAPPING
+        self._save_stage()
```
mgfritch, line 147: (nit) make this a setter on the bootrapping property? or create a generic `set_stage` func?

src/gravel/controllers/nodes/deployment.py
```diff
@@ -13,10 +13,34 @@
...
 from gravel.controllers.gstate import GlobalState
+from gravel.controllers.nodes.conn import ConnMgr
+
+from gravel.controllers.nodes.errors import (
+    NodeAlreadyJoiningError,
+    NodeBootstrappingError,
+    NodeHasBeenDeployedError,
+    NodeHasJoinedError
+)
+from gravel.controllers.nodes.messages import (
```
mgfritch, line 43: (nit) rename this `gravel.controllers.nodes.models` ?

src/gravel/controllers/nodes/deployment.py
```diff
@@ -124,3 +148,120 @@ def mark_ready(self) -> None:
...
+            await conn.close()
+            self._state.mark_error()
+            return False
+
+        assert reply.type == MessageTypeEnum.WELCOME
+        welcome = WelcomeMessageModel.parse_obj(reply.data)
+        assert welcome.pubkey
+        assert welcome.cephconf
+        assert welcome.keyring
+        assert welcome.etcd_peer
```
mgfritch, line 250: wouldn't pydantic raise during `parse_obj` for these?

src/gravel/controllers/nodes/deployment.py
```diff
@@ -124,3 +148,120 @@ def mark_ready(self) -> None:
...
+            logger.debug(f"join > wrote pubkey to {authorized_keys}")
+
+        cephconf_path: Path = Path("/etc/ceph/ceph.conf")
+        keyring_path: Path = Path("/etc/ceph/ceph.client.admin.keyring")
+        if not cephconf_path.parent.exists():
+            cephconf_path.parent.mkdir(0o755)
+        cephconf_path.write_text(welcome.cephconf)
+        keyring_path.write_text(welcome.keyring)
+        keyring_path.chmod(0o600)
+        cephconf_path.chmod(0o644)
```
mgfritch, line 278: (nit) wondering if this belongs in a Config class (or similar)?

mgfritch (approved):
couple of suggestions, but generally looks good!

src/gravel/controllers/nodes/deployment.py
jecluis: thought about properties, but wanted to make it as explicit as possible for code readability. as for a generic, I really didn't want to expose the enums to the caller.

src/gravel/controllers/nodes/deployment.py
jecluis: well, they are models, true, but they are used as messages. I think for clarity it's better to have that as 'messages' and, eventually, move common models to a 'models' file.

src/gravel/controllers/nodes/deployment.py
jecluis: it would, but given they're strings, it wouldn't catch an empty string.

src/gravel/controllers/nodes/deployment.py
jecluis: it probably belongs somewhere else, but I don't think it would be in the Config class. Then again, I don't expect this to be used anywhere else, so I'm not particularly bothered by having it here :shrug:

PR #464, opened 2021-05-02
images: bundle containers

### LINKED ISSUE #457: Include containers necessary for initial deployment inside Aquarium image
When deploying Ceph on each node[1], the ceph container image is pulled from (at the time of writing) docker.io/ceph/ceph:v16. That takes a certain amount of time (and also an internet connection).

Once #449 is merged, we'll also be pulling quay.io/coreos/etcd:latest, and that presumably needs to complete before Aquarium can actually be used at all.

I assume there may be other containers in future too.

So, I had a crazy idea. Can we do a `podman pull` of all necessary containers at image build time, so that they are present and ready to use immediately inside Aquarium?

[1] Actually, it looks like the ceph container is pulled automagically at runtime _before_ deployment, I assume as a result of Aquarium periodically running `cephadm gather-facts` and/or `cephadm inventory`, but the point still stands, this requires a certain amount of time, and an internet connection.

jecluis:
Resolves #457 

### DIFF
--- a/images/microOS/disk.sh
+++ b/images/microOS/disk.sh
@@ -37,11 +37,13 @@ mount -t tmpfs none /dev/shm
 mount -t proc none /proc
 mount -t tmpfs none /run
 
-/usr/bin/podman pull quay.io/coreos/etcd:latest
+# setting "--events-backend none" means podman doesn't try
+# (and fail) to log a "system refresh" event to the journal
+/usr/bin/podman --events-backend none pull quay.io/coreos/etcd:latest
 # we don't get to use cephadm directly because it will
 # try running a container inside the chroot, and that
 # fails with a bang.
-/usr/bin/podman pull docker.io/ceph/ceph:v16
+/usr/bin/podman --events-backend none pull docker.io/ceph/ceph:v16
 
 # cleanup
 umount /run
--- a/images/microOS/disk.sh
+++ b/images/microOS/disk.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+#
+# Aquarium image kiwi scripts
+# Copyright (C) 2021 SUSE LLC
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+set -euxo pipefail
+
+if ! /usr/bin/podman --version >/dev/null ; then
+  echo "ERROR: can't find podman"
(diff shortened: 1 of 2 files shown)


jhesketh (commented):
Unless I'm missing something, this script isn't actually ran, so we're not including the images so much as a helper for them.

jecluis:
On Mon, May 3, 2021, 04:46 Joshua Hesketh ***@***.***> wrote:

> ***@***.**** commented on this pull request.
>
> Unless I'm missing something, this script isn't actually ran, so we're not
> including the images so much as a helper for them.
>
Much like config.sh, disk.sh is run at a given point in the kiwi workflow.
In this case, disk.sh is called on the final image instead of the working
filesystem tree.

So no, it's not a helper, and it is really including the images in the
build.

jhesketh:
> On Mon, May 3, 2021, 04:46 Joshua Hesketh ***@***.***> wrote: ***@***.**** commented on this pull request. Unless I'm missing something, this script isn't actually ran, so we're not including the images so much as a helper for them.
> Much like config.sh, disk.sh is run at a given point in the kiwi workflow. In this case, disk.sh is called on the final image instead of the working filesystem tree. So no, it's not a helper, and it is really including the images in the build.

Ah, okay, I had wondered if something like that was the case. (I'm not very familiar with kiwi). :+1:

images/microOS/disk.sh
```diff
@@ -0,0 +1,54 @@
...
+echo "nameserver 1.1.1.1" > /etc/resolv.conf
+
+# and we need a basic running system too
+mount -t sysfs none /sys
+mount -t devtmpfs none /dev
+mount -t tmpfs none /dev/shm
+mount -t proc none /proc
+mount -t tmpfs none /run
+
+/usr/bin/podman pull quay.io/coreos/etcd:latest
```
tserong, line 40:
```suggestion
# setting "--events-backend none" means podman doesn't try
# (and fail) to log a "system refresh" event to the journal
/usr/bin/podman --events-backend none pull quay.io/coreos/etcd:latest
```

images/microOS/disk.sh
```diff
@@ -0,0 +1,54 @@
...
+# we don't get to use cephadm directly because it will
+# try running a container inside the chroot, and that
+# fails with a bang.
+/usr/bin/podman pull docker.io/ceph/ceph:v16
```
tserong, line 44:
```suggestion
/usr/bin/podman --events-backend none pull docker.io/ceph/ceph:v16
```

tserong (approved):
Works like a charm, although I noticed this ugly thing in the middle of the disk.sh output:

```
time="2021-05-03T07:39:44Z" level=error msg="unable to write system event: \"write unixgram @00148->/run/systemd/journal/socket: sendmsg: no such file or directory\""
```

Turns out podman tries to log a "system refresh" event to the journal. In the context of this script, I don't think we care at all about podman events, so I've suggested adding `--events-backend none` to each of the `podman pull` invocations to get rid of this.

--events-backend none

tserong (approved):

PR #467, opened 2021-05-03
glass: finish deployment on wizard's end

jecluis:


votdev:
Sorry, didn't realize that you already opened a PR to fix that issue. My PR marks the bootstrapping as finished at an earlier stage (like the old one). So we have to decide which one we should use. See https://github.com/aquarist-labs/aquarium/pull/468

jecluis:
Lets merge yours and close this one :)

PR #476, opened 2021-05-05
gravel: fix inventory gathering, and speed up device probing

jecluis:

### DIFF
--- a/src/gravel/controllers/resources/inventory_sub.py
+++ b/src/gravel/controllers/resources/inventory_sub.py
@@ -1,0 +1,32 @@
+# project aquarium's backend
+# Copyright (C) 2021 SUSE, LLC.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+from __future__ import annotations
+from typing import Awaitable, Callable
+from gravel.cephadm.models import NodeInfoModel
+
+
+class Subscriber:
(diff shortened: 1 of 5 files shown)


src/gravel/controllers/resources/inventory_sub.py
```diff
@@ -0,0 +1,35 @@
...
+        cb: Callable[[NodeInfoModel], Awaitable[None]],
+        once: bool,
+        registry: List[Subscriber]
+    ):
+        self.cb = cb
+        self.once = once
+        self._registry = registry
+
+    def unsubscribe(self) -> None:
+        self._registry.remove(self)
```
jhesketh, line 35: Seems odd for Inventory to add the subscriber to the inventory but of the subscriber itself to unsubscribe. I think NodeMgr should call back to the Inventory to remove any subs.

jhesketh (changes requested):
Perhaps a nit.

Also, are we ready to require tests going forward?

jecluis:
@jhesketh addressed, repushed. no tests though, I'll have to figure out how to write those and, tbh, I'd rather have this in soon instead of waiting for them.

jhesketh (changes requested):
The change looks good, but we need to choose here how strict we want to be with requiring tests going forward.

From our discussions it is my impression that we want to require at least some level of testing. This does mean things will move slower and things may be held up more, but I also believe that is the only way we're going to make any progress on growing our test coverage. As such I'm marking this as a "requests changes", but should others disagree the change still looks good and I'm happy for it to merge.

src/gravel/tests/unit/controllers/resources/test_inventory.py
```diff
@@ -0,0 +1,92 @@
...
+    mocker: MockerFixture,
+    gstate: GlobalState
+):
+
+    nodeinfo: NodeInfoModel = NodeInfoModel(
+        hostname="foo",
+        model="",
+        vendor="",
+        kernel="123.4",
+        operating_system="fancything",
```
mgfritch, line 38: fancy!

mgfritch (approved):
lgtm!

jhesketh (commented):
Awesome, thanks for adding a test :-)

PR #460, opened 2021-04-30
Fixtures

jhesketh:
Rework how we make external connections to enable easier testing.

This means we pass objects between methods to keep the connections established (rather than rebuilding various connections on requests).

This should ease testing allowing us to stub boundary layers.

PR #489, opened 2021-05-09
aqrtest: functional tests battery for aquarium's backend

jecluis:
Introduce a framework to test a live aquarium deployment.

Tests are meant to be simple to write, but this likely can be further enhanced.

Each test runs a brand new aquarium instance, to avoid tainting internal state in case of exceptions.

What it doesn't do however is actually deploying VMs to run the tests on. Which it needs to, because some tests will eventually taint the deployment irreversibly (e.g., bootstrapping a cluster will leave deployed containers and assimilated disks behind). This is future looking improvements.

The work in this PR has explicitly added a test that is known to fail, as an example of how things can be broken. Said failure needs to be fixed before enabling PR checking using aqrtest.

### DIFF
--- a/images/microOS/root/usr/lib/systemd/system/aquarium.service
+++ b/images/microOS/root/usr/lib/systemd/system/aquarium.service
@@ -5,7 +5,7 @@ After=network.target
 [Service]
 User=root
 WorkingDirectory=/usr/share/aquarium
-ExecStart=/usr/local/bin/uvicorn aquarium:app --host 0.0.0.0 --port 1337
+ExecStart=/usr/local/bin/uvicorn aquarium:app_factory --factory --host 0.0.0.0 --port 1337
 Restart=always
 
 [Install]
--- a/run_aquarium.sh
+++ b/run_aquarium.sh
@@ -72,7 +72,7 @@ pushd ${SCRIPT_DIR}/src &>/dev/null
 $is_debug && export AQUARIUM_DEBUG=1
 $has_config && export AQUARIUM_CONFIG_DIR=${config_path}
 
-uvicorn aquarium:app --host 0.0.0.0 --port ${port}
+uvicorn aquarium:app_factory --factory --host 0.0.0.0 --port ${port}
 
 popd &>/dev/null
 
--- a/src/aquarium.py
+++ b/src/aquarium.py
@@ -19,6 +19,7 @@ import os
 from fastapi import FastAPI
 from fastapi.staticfiles import StaticFiles
 from fastapi.logger import logger as fastapi_logger
+import uvicorn
 
 from gravel.controllers.gstate import GlobalState, setup_logging
 from gravel.controllers.nodes.mgr import NodeMgr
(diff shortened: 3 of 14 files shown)


jhesketh:
So what is the advantage of this approach over #460?

Can we at least get 460 merged first or rebase this on that branch please?

We should perhaps have a discussion here about the direction because I think there is some duplication of efforts.

jecluis:
> So what is the advantage of this approach over #460?

I see them as complementary: these are actual functional tests, that do test the whole application stack; the others are unit tests, meant to validate what we believe are the correct behaviors of those paths.

The obvious advantage of these over the other is that we're actively testing the application's behavior from the consumer's point of view, as a live system, which we can't do when we mock half of the application. An example of why these complement the other is exactly the failures on #460 , where all the unit tests are passing but the application doesn't actually work.

> Can we at least get 460 merged first or rebase this on that branch please?

I'm reusing some of your commits exactly with that in mind, although not rebasing on top until the failures are fixed.


> We should perhaps have a discussion here about the direction because I think there is some duplication of efforts.

I don't see a duplication of efforts. As I said, I see them as complementary.

jhesketh:
> > So what is the advantage of this approach over #460?
> 
> I see them as complementary: these are actual functional tests, that do test the whole application stack; the others are unit tests, meant to validate what we believe are the correct behaviors of those paths.

The refactor and fixtures worked in 460 are to allow exactly this. They end up supplying an endpoint that you exercise the whole stack.

> 
> The obvious advantage of these over the other is that we're actively testing the application's behavior from the consumer's point of view, as a live system, which we can't do when we mock half of the application. An example of why these complement the other is exactly the failures on #460 , where all the unit tests are passing but the application doesn't actually work.

The only parts that are mocked are calls to other applications that don't exist in pytest (like cephadm or etcd).

The failures need fixing, and that's on the top of my todo for Monday morning.

We can add better exception handling to get failures from uvicorn.

> 
> > Can we at least get 460 merged first or rebase this on that branch please?
> 
> I'm reusing some of your commits exactly with that in mind, although not rebasing on top until the failures are fixed.

Merging something like this will create a significant amount more work for 460. Fair enough if we don't think that it's going to merge, but I would rather we work together in getting that in before creating such work.

> 
> > We should perhaps have a discussion here about the direction because I think there is some duplication of efforts.
> 
> I don't see a duplication of efforts. As I said, I see them as complementary.

I'm not sure I agree. I think what you're doing here can be achieved with the fixtures in 460 (and was entirely the point).

jecluis:
> > > So what is the advantage of this approach over #460?
> > 
> > 
> > I see them as complementary: these are actual functional tests, that do test the whole application stack; the others are unit tests, meant to validate what we believe are the correct behaviors of those paths.
> 
> The refactor and fixtures worked in 460 are to allow exactly this. They end up supplying an endpoint that you exercise the whole stack.
> 
> > The obvious advantage of these over the other is that we're actively testing the application's behavior from the consumer's point of view, as a live system, which we can't do when we mock half of the application. An example of why these complement the other is exactly the failures on #460 , where all the unit tests are passing but the application doesn't actually work.
> 
> The only parts that are mocked are calls to other applications that don't exist in pytest (like cephadm or etcd).

This is crucial to test though. We can't validate Aquarium without validating that what we are doing makes sense if we are ignoring two (three, if we count librados) of the most important parts of the whole stack.


> The failures need fixing, and that's on the top of my todo for Monday morning.
> 
> We can add better exception handling to get failures from uvicorn.

:+1: 


> > > Can we at least get 460 merged first or rebase this on that branch please?
> > 
> > 
> > I'm reusing some of your commits exactly with that in mind, although not rebasing on top until the failures are fixed.
> 
> Merging something like this will create a significant amount more work for 460. Fair enough if we don't think that it's going to merge, but I would rather we work together in getting that in before creating such work.

Hence why the plan has been merging #460 first and then merging this one. As I said, the commits are here exactly with that in mind, so that I don't have to go through a bunch of trouble rebasing on top of an up to date main.

> > > We should perhaps have a discussion here about the direction because I think there is some duplication of efforts.
> > 
> > 
> > I don't see a duplication of efforts. As I said, I see them as complementary.
> 
> I'm not sure I agree. I think what you're doing here can be achieved with the fixtures in 460 (and was entirely the point).

As said above, we are not doing the same thing if underneath we are mocking things that are crucial, like librados, etcd, and cephadm, as well as their replies. That makes sense in a context; this is not the context I'm trying to target.

jecluis:
> Okay then, I may have misunderstood sorry.
> 
> So the part I'm missing here then is how tests are expected to interact with cephadm/librados etc.?
> 
> Is the intend for these _not_ to be run by pytest?
> 
> Would these be ran in an expected environment (ie, one that is prepared by wasser)?
> 
> What happens if the environment differs from what is expected etc, and how are developers meant to run/validate them?

These will be run inside our VM deployment, at this point by magic (i.e., the user, later by our testing infra). It's expected the environment is known. At the moment, expecting 4 disks, one of which reserved for the OS.

So, pytest is not involved. So much so that I have explicitly ignored this directory in `tox.ini`.

Regardless, I'm not particularly happy with them as it is yet, because it works fine for single-node deployment testing, but won't be easy for multiple node testing.

jhesketh:
> > Okay then, I may have misunderstood sorry.
> > So the part I'm missing here then is how tests are expected to interact with cephadm/librados etc.?
> > Is the intend for these _not_ to be run by pytest?
> > Would these be ran in an expected environment (ie, one that is prepared by wasser)?
> > What happens if the environment differs from what is expected etc, and how are developers meant to run/validate them?
> 
> These will be run inside our VM deployment, at this point by magic (i.e., the user, later by our testing infra). It's expected the environment is known. At the moment, expecting 4 disks, one of which reserved for the OS.
> 
> So, pytest is not involved. So much so that I have explicitly ignored this directory in `tox.ini`.
> 
> Regardless, I'm not particularly happy with them as it is yet, because it works fine for single-node deployment testing, but won't be easy for multiple node testing.

Okay, that's where my confusion was.

That makes more sense. I was intending on doing this sort of testing down the road, envisioning that it would be set up with wasser (and likely somewhat tightly coupled to it to ensure a consistent environment).

To make things clearer I think it'd be good for these tests to be outside of the `gravel/tests/` directory. Perhaps with a README on how to run them since their entrypoint is different to the standard pytest. (and maybe for that matter not prefixing files with test*.py since that is an expected unittest format).

jecluis:
closing this as I'm working on something that makes a little bit more sense

PR #452, opened 2021-04-27
[tools]: [Vagrant setup include VirtualBox image type]

AvengerMoJo:
Allow auto detect image type to include provider e.g. libvirt
Or if the image type is VirtualBox it will use VirtualBox syntax

Signed-off-by: Alex Lau (AvengerMoJo) <alau@suse.com>

## Checklist
- [ ] References issues, create if required
- [*] Updates documentation if necessary
- [ ] Includes tests for new functionality or reproducer for bug

PR #492, opened 2021-05-11
aqrdev: rework internal structure

jecluis:
This patchset focus on reworking `aqrdev`'s internal organization, splitting off auxiliary functions into a separate module to then be consumed by upcoming work, reducing the amount of code in the one single file we had.

### DIFF
--- a/tools/aqrdev
+++ b/tools/aqrdev
@@ -13,150 +13,96 @@
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU General Public License for more details.
 
-import os
 import sys
 import errno
 import json
 from typing import Any, Dict, List, Optional, Tuple
 import click
 import logging
-import subprocess
-import shlex
-import shutil
-import random
 from click.decorators import make_pass_decorator
 from pathlib import Path
-from datetime import datetime as dt
-from pydantic import BaseModel
+
+from libaqr.deployment import Deployment, DeploymentModel, get_deployments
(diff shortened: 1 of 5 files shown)


mgfritch (approved):

AvengerMoJo:
Just to say in my PR #452 the concept of "Provider" for vagrant is added and it will be still missing in the new structure. If the user using vagrant not only using libvirt then it may create some issue.

jecluis:
> Just to say in my PR #452 the concept of "Provider" for vagrant is added and it will be still missing in the new structure. If the user using vagrant not only using libvirt then it may create some issue.

No worries. I'll rebase your PR and fix the missing pieces.

tools/aqrdev
```diff
@@ -852,8 +852,14 @@ def cmd_stop(ctx: AppCtx, name: str) -> None:
 @app.command("shell")
 @click.argument("name", required=True, type=str)
 @click.option("-n", "--node", required=False, type=str)
+@click.option("-c", "--command", required=False, type=str)
```
jhesketh, line 363: (nit) We should pass in a help string

jhesketh (approved):
Looks good, although it is difficult to review such a large change so I am mostly relying on the successful wasser build and the fact that this is a refactor rather than all new code.

tools/aqrdev
jecluis: Agreed. This should be done for pretty much all of the commands :\

PR #494, opened 2021-05-12
build: Fix opensuse builds on cloud images

jhesketh:
When doing a build on an opensuse tumbleweed cloud image, the locale
isn't already set correctly for kiwi (see
https://www.suse.com/support/kb/doc/?id=000019958).

kiwi also needed dosfstools at some point. I'm not sure why, but
installing it unblocked me.

### DIFF
--- a/tools/build-image.sh
+++ b/tools/build-image.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+LC_ALL=C.UTF-8
+
 rootdir=
 
 find_root() {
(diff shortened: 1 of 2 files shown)


tools/build-image.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+LC_ALL=C.UTF-8
```
kshtsk, line 3: This looks familiar to me, when I use tools on MacOS and trying to ssh to any linux box.

jecluis (approved):
haven't tested, but looks generally okay to me, and the build seems to be passing, so should be okay to merge.

PR #495, opened 2021-05-12
gravel/tests: add missing type hints

jecluis:

### DIFF
--- a/src/gravel/tests/conftest.py
+++ b/src/gravel/tests/conftest.py
@@ -13,15 +13,21 @@
 
 import os
 from types import SimpleNamespace
-from typing import Optional
 import pytest
 import sys
 from pyfakefs import fake_filesystem  # pyright: reportMissingTypeStubs=false
 from _pytest.fixtures import SubRequest
 from typing import (
     Callable,
-    Tuple
+    Tuple,
+    Any,
+    Awaitable,
+    Dict,
+    Optional,
+    cast
 )
+from pytest_mock import MockerFixture
+from fastapi import FastAPI
 
 from gravel.controllers.gstate import GlobalState
 from gravel.controllers.kv import KV
@@ -33,7 +39,7 @@ import logging
 DATA_DIR = os.path.join(os.path.dirname(__file__), 'data')
 
 
-def mock_ceph_modules(mocker):
+def mock_ceph_modules(mocker: MockerFixture):
     class MockRadosError(Exception):
         def __init__(self, message: str, errno: Optional[int] = None):
             super(MockRadosError, self).__init__(message)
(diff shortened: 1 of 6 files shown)


src/gravel/tests/conftest.py
```diff
@@ -85,7 +91,9 @@ def ceph_conf_file_fs(
 
 
 @pytest.fixture()
-def get_data_contents(fs: fake_filesystem.FakeFilesystem):
+def get_data_contents(  # type: ignore
+    fs: fake_filesystem.FakeFilesystem
+) -> str:
```
mgfritch, line 96:
```suggestion
def get_data_contents(
    fs: fake_filesystem.FakeFilesystem
) -> Generator[Callable[[str, str], str], None, None]:
```

these ones are fun, but I think the above is correct

src/gravel/tests/conftest.py
```diff
@@ -85,7 +91,9 @@ def ceph_conf_file_fs(
...
     def _get_data_contents(dir: str, fn: str):
```
mgfritch, line 97:
```suggestion
    def _get_data_contents(dir: str, fn: str) -> str:
```

src/gravel/tests/conftest.py
```diff
@@ -99,10 +107,10 @@ def _get_data_contents(dir: str, fn: str):
         with open(os.path.join(dir, fn), 'r') as f:
             contents = f.read()
         return contents
-    yield _get_data_contents
+    yield _get_data_contents  # type: ignore
```
mgfritch, line 110:
```suggestion
    yield _get_data_contents
```

mgfritch (approved):
an improvement overall, but it would be nice to avoid adding more `type: ignore` ..

src/gravel/tests/conftest.py
jhesketh: I think you're correct, but eww :-p

jhesketh (approved):
lgtm! Thanks for that.

(Not sure if you want to fix up Mike's suggestion first or not)

src/gravel/tests/conftest.py
jecluis: tried that. It works to shut up mypy, but won't shut up pyright. That's because when it's called, it's not called as a generator, butt as a function (because pytest is doing magic in the background). So pyright will be annoyed because we're calling on a generator without actually using it as a generator.

TL;DR: pytest typing support sucks, and this was the closest I could get to make that function work.

src/gravel/tests/conftest.py
jecluis: This works if we're using the generator return, but not in any other way. While we get correctness for this function, we lose correctness (at least from the expected behavior point of view) for the rest of the tests that are using this function. We can't be telling the type checker we're yielding a function here and expect `get_data_contents` to be returning a string without calling on `next()`.

jecluis:
Thanks. I'm kicking Mike's comments down the road to a time when someone (either me or whomever) has time to figure out how to type check pytest generators correctly. I actually started my endeavor from where Mike's comments are, and after an afternoon of insanity I ended up choosing to ignore types for that function to make the rest work.

PR #496, opened 2021-05-13
docs/project-plan: add roadmap and updated milestones

jecluis:

### DIFF
--- a/doc/project-plan/project-plan.rst
+++ b/doc/project-plan/project-plan.rst
@@ -85,6 +85,10 @@ Identify risks and ensure team members monitor the risks associated with them.
 Milestones
 ~~~~~~~~~~
 
+The following were the initially planned milestones, meant to guide us for the
+first three months of the project. For up to date, forward looking milestones,
+see :ref:`aquarium-roadmap`.
+
 .. list-table::
    :widths: 35 35 25
    :header-rows: 1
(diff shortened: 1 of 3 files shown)


doc/project-plan/project-plan.rst
```diff
@@ -107,6 +107,34 @@ Milestones
      - March 31, 2021
    * - Milestone 3
      -
+         - Install Wizard improvements
+         - Increase testing coverage
+     - May 7, 2021
+   * - Milestone 4
```
asettle, line 113: So, I don't think we need to keep adding this to the project plan. I think we can probably just put something up the top that says this was for the first 3 months (3 milestones) and use the roadmap below to track the rest. Otherwise all we're doing is keeping track of this in about 40 places. Also, each milestone should have a definition when we create them...

asettle (commented):
Suggestion inline

doc/project-plan/project-plan.rst
jecluis: good point. Will adjust.

jecluis:
@asettle updated. let me know what you think.

holgis (approved):

asettle (approved):
LGTM

votdev (approved):
LGTM

PR #499, opened 2021-05-14
aqrtest: functional test battery against a live cluster

jecluis:
With this we are introducing a new script, `aqrtest.py`, which will run tests within the `tests/suites/` directory against a live vagrant cluster.

Tests are organized in Test Cases, which can be one or contain multiple Case Units. Each Test Case will declare its requirements, based on which the cluster will be deployed.

~This is still on-going work, and~ I'm looking for feedback.

**Note:** even after merge, we can't immediately make this the default on all PRs because it is expected that a lot of the API tests are going to fail. The API needs to be fixed before we can enable automatic checks as merging criteria.

### DIFF
--- a/tests/suites/smoke/test_api_bootstrap.py
+++ b/tests/suites/smoke/test_api_bootstrap.py
@@ -11,6 +11,23 @@
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU General Public License for more details.
 
-import libaqr.vagrant
 
-print(libaqr.vagrant.Vagrant.box_list())
+from libaqr.testing import TestCase, caseunit, requirements
+
+
+@requirements(
+    disks=4,
+    nodes=1,
+    nics=1
+)
+class TestAPIBootstrap(TestCase):
+
+    @caseunit
+    async def test_func_1(self):
+        print("running test func 1")
+        assert True
+
+    @caseunit
+    async def test_func_2(self):
+        print("running test func 2")
+        assert True
--- a/tests/suites/smoke/test_api_bootstrap.py
+++ b/tests/suites/smoke/test_api_bootstrap.py
@@ -1,0 +1,16 @@
+# project aquarium's testing battery
+# Copyright (C) 2021 SUSE, LLC.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+import libaqr.vagrant
+
+print(libaqr.vagrant.Vagrant.box_list())
--- a/tools/aqrtest.py
+++ b/tools/aqrtest.py
@@ -63,6 +63,9 @@ def cmd_run(
         click.secho("No suites found", fg="yellow")
         sys.exit(0)
 
+    # add suites parent directory to pythonpath so we can import modules
+    sys.path.append(suitesdir.parent.as_posix())
+
     # prepare image
     try:
         Image.add(find_builds_path(), image)
--- a/tools/aqrtest.py
+++ b/tools/aqrtest.py
@@ -17,13 +17,14 @@ import errno
 from pathlib import Path
 from libaqr.errors import BoxAlreadyExistsError, ImageNotFoundError, VagrantError
 from libaqr.images import Image
-from libaqr.runner import Runner
+from libaqr.runner import Runner, RunnerResult
 from libaqr.suites import (
+    SuiteEntry,
     get_available_suites,
     get_suite_tests
 )
 from libaqr.misc import find_builds_path
-from typing import Optional
+from typing import List, Optional, Tuple
 import click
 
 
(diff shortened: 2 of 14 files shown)


tserong (commented):
Overall looks like a pretty neat framework to me. As discussed on Slack just now the first set of tests don't quite do the right thing (they fail due to a timeout), but that's not the fault of the framework. This PR could do with at least one more set of eyes though ;-)

tests/suites/smoke/test_api_bootstrap.py
```diff
@@ -0,0 +1,16 @@
+# project aquarium's testing battery
```
jhesketh, line 1: It's totally a nit, but I'd personally like to avoid naming files in the unit-test format of `test_*.py` to avoid confusion with test runners. Perhaps `aqrtest_*.py`?

tools/aqrtest.py
```diff
@@ -0,0 +1,120 @@
...
+    suite: Optional[str],
+    suites_path: Optional[Path]
+) -> None:
+    """ Run a test or a suite against a given image """
+
+    suitesdir = suites_path if suites_path else Path("tests/suites")
+
+    if not suitesdir.exists():
+        click.secho("error: unable to find suites directory", fg="red")
+        sys.exit(errno.ENOENT)
```
jhesketh, line 53: I think we've come across this before. Wouldn't it be better to use click's exit handling: https://click.palletsprojects.com/en/8.0.x/api/#click.Context.exit

This will also allow us to exit by raising appropriate exceptions etc.

tools/aqrtest.py
```diff
@@ -0,0 +1,120 @@
...
+@click.option("--suites-path", required=False, type=Path)
+def cmd_run(
+    image: str,
+    test: Optional[str],
+    suite: Optional[str],
+    suites_path: Optional[Path]
+) -> None:
+    """ Run a test or a suite against a given image """
+
+    suitesdir = suites_path if suites_path else Path("tests/suites")
```
jhesketh, line 49: (nit) could search both pwd and relative to this aqrtest.py file.

tools/aqrtest.py
```diff
@@ -0,0 +1,120 @@
...
+    if not suitesdir.exists():
+        click.secho("error: unable to find suites directory", fg="red")
+        sys.exit(errno.ENOENT)
+
+    available_suites = get_available_suites(suitesdir)
+    if suite:
+        if suite not in available_suites:
+            click.secho(f"error: unknown suite '{suite}'", fg="red")
+            sys.exit(errno.EINVAL)
+        available_suites = [suite]
```
jhesketh, line 60: (nit) Only one test and suite can be supplied, but if they aren't supplied then all available_suites's are used. Would be useful to allow multiple.

tools/aqrtest.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+@click.group()
+def app() -> None:
+    pass
+
+
+@app.command("run")
+@click.argument("image", required=True, type=str)
+@click.option("-t", "--test", required=False, type=str)
```
jhesketh, line 38: If I'm reading this correctly, the test name can include the suite (eg `suite/test`), so the suite option is probably redundant. I'd suggest dropping it if we're going to support multiple tests on one `run` cmd, and also to avoid confusion (since we've gone to the effort of defining a syntax etc).

tools/aqrtest.py
```diff
@@ -0,0 +1,120 @@
...
+        # we are going to reuse that box
+        pass
+
+    deployments_path = Path("aqrtest-deployments")
+    deployments_path.mkdir(exist_ok=True)
+
+    for entry in get_suite_tests(suitesdir, suite, test):
+        runner = Runner(image, entry)
+        runner.setup(deployments_path)
+        success = runner.run()
```
jhesketh: Should put this in a `try` statement to do the teardown on exception etc.
Perhaps also with the setup if there are teardown steps that can still be performed.

tools/libaqr/runner.py
```diff
@@ -0,0 +1,125 @@
...
+            from datetime import datetime as dt
+            now = dt.now().isoformat().replace(":", "").split(".")[0]
+            return f"aqrtest-{self._test.name}-{now}"
+
+        self._run_name = _gen_name()
+        try:
+            self._deployment = Deployment.create(
+                name=self._run_name,
+                box=self._boxname,
+                num_nodes=2, num_disks=4, num_nics=1,
```
jhesketh: Shouldn't this be declarable by the test itself?

tools/libaqr/testing.py
```diff
@@ -0,0 +1,152 @@
...
+class Requirements(BaseModel):
+    disks: int = Field(4, description="number of disks")
+    nics: int = Field(1, description="number of network interfaces")
+    nodes: int = Field(2, description="number of nodes")
+
+
+UnitCaseType = Callable[..., Awaitable[None]]
+
+
+class TestCase:
```
jhesketh, line 30: (nit) Similarly would like this to be `AqrtestCase` or something.

jhesketh (approved):
This is pretty awesome start. All of my comments are either nits or things that can be improved in subsequent changes. Since this PR doesn't affect existing coverage or code etc, I'm happy to see it merged and then followed up on (since reviewing is already unwieldy).

tools/libaqr/runner.py
jhesketh: This was done in a subsequent commit already. (Sorry, I review commit-by-commit).

tools/aqrtest.py
jecluis: This comment is on outdated code.

jecluis:
Alright, let's address those things in another pass, including file rename etc.

PR #502, opened 2021-05-17
Use python black

jhesketh:
This change is purely to do with formatting and also likely purely subjective.

However, there have been a few times in reviews that people have had different styles and that has caused either
confusion or a desire to consolidate.

Black takes all these arguments away by being the dictator in what is correct. Read more about it here:
https://black.readthedocs.io/en/stable/index.html

As this is very subjective, we should make sure people are onboard with this styling before merging it.


src/aquarium.py
```diff
@@ -69,28 +69,28 @@ async def aquarium_startup(aquarium_app: FastAPI, aquarium_api: FastAPI):
 
     # Set up all of the tickers
     devices: Devices = Devices(
-        gstate.config.options.devices.probe_interval,
-        nodemgr,
-        ceph_mgr,
-        ceph_mon
+        gstate.config.options.devices.probe_interval, nodemgr, ceph_mgr, ceph_mon
```
jecluis, line 72: gotta say, this annoys me a lot. I think it makes readability worse.

src/aquarium.py
```diff
@@ -109,49 +109,45 @@ async def aquarium_shutdown(aquarium_app: FastAPI, aquarium_api: FastAPI):
 
 
 def aquarium_factory(
-    startup_method=aquarium_startup,
-    shutdown_method=aquarium_shutdown,
-    static_dir=None
+    startup_method=aquarium_startup, shutdown_method=aquarium_shutdown, static_dir=None
```
jecluis, line 112: and this also makes readability worse than having a giant line with all the options there.

src/pyproject.toml
```diff
@@ -0,0 +1,14 @@
+[tool.black]
+line-length = 88
```
jecluis, line 2: can we do 80-81? I admit it's mostly for my benefit, because

image

jecluis (commented):
I have some issues with the formatting, but it is generally okay IMO.

The thing that comes to mind though is: why is this better than flake8?

src/pyproject.toml
jecluis: okay, in retrospect, I think I can live with the line wrapping. Mostly because I think `black` will do the right thing when it comes to it:

image

src/aquarium.py
jecluis: in retrospect, I think I can live with this sort of line wrapping. It's not what I'd prefer, but in the end `black` seems to be doing the right thing if even that line is too big:

```
# in:

def very_important_function(template: str, *variables, file: os.PathLike, engine: str, header: bool = True, debug: bool = False):
    """Applies `variables` to the `template` and writes to `file`."""
    with open(file, 'w') as f:
        ...

# out:

def very_important_function(
    template: str,
    *variables,
    file: os.PathLike,
    engine: str,
    header: bool = True,
    debug: bool = False,
):
    """Applies `variables` to the `template` and writes to `file`."""
    with open(file, "w") as f:
        ...

```

votdev:
> I have some issues with the formatting, but it is generally okay IMO.
> 
> The thing that comes to mind though is: why is this better than flake8?

I'm in favor of autopep8 because black has too much options IMO.

src/aquarium.py
jhesketh: I don't disagree. There are some styles of black that don't read as well to me, but I'm happy with that trade off for otherwise strict and reasonably sensible enforcement.

src/aquarium.py
jhesketh: We could lower the line length which might solve this.

src/pyproject.toml
jhesketh: While I'm fine with 81, you seem to have plenty of blank space in those files?

jhesketh:
> > I have some issues with the formatting, but it is generally okay IMO.
> > The thing that comes to mind though is: why is this better than flake8?
> 
> I'm in favor of autopep8 because black has too much options IMO but i see that black is part of PSF, so maybe that's a serious argument.

Actually black doesn't have many options at all - that's kinda the point. However, it does have very many opinions.

The main difference is that pep8 allows for multiple ways to do something, whereas black enforces just one way. The idea is to remove arguments and to be consistent above all else (even at the potential cost of readability). black also has a very good auto-formatter that ensures the python byte-code is unchanged.

It's certainly opinionated and divisive, but once we have agreement to use it then it leaves no room for further arguments. I do admit it's not necessarily right in all cases or for us, but from seeing some of the inconsistencies we have and nits from reviews I decided it might be helpful.

src/aquarium.py
mgfritch: I'd rather not argue about line length ;)

mgfritch (commented):
wait a second... I thought [orange is the new black](https://pypi.org/project/oitnb/)?!? - but really, I'm not a fan of any of the formatters (autopep8 included).

I would personally prefer a good style guide with a linter (e.g. flake8) and leave the remaining format/readability up to the style of the developer.

> I have some issues with the formatting, but it is generally okay IMO.

same, I find the formatting to be more rigid than necessary and occasionally will make the readability worse. 

> The thing that comes to mind though is: why is this better than flake8?

exactly, in my opinion we _could_ use black as a formatter, but we _should_ likely still use flake8 as a linter.

src/pyproject.toml
jecluis: because those lines are not close to 80 cols. ;)

jhesketh:
Should we do a vote on this next meeting or do people feel strongly enough that this is annoying and we shouldn't bother?

jecluis:
Yeah, let's talk about this on our next meeting, and figure out what the concerns are and whether they are surmountable.

asettle:
> Yeah, let's talk about this on our next meeting, and figure out what the concerns are and whether they are surmountable.

Did we have this conversation? @jhesketh @jecluis @mgfritch ?

jecluis:
> > Yeah, let's talk about this on our next meeting, and figure out what the concerns are and whether they are surmountable.
> 
> Did we have this conversation? @jhesketh @jecluis @mgfritch ?

We did, and we've postponed it to a later date when at least one of us is annoyed enough to pursue this avenue. I think it's okay to close the PR though, given it's full of conflicts by now and will need to be redone anyway should we move forward with black at some point.

asettle:
> > > Yeah, let's talk about this on our next meeting, and figure out what the concerns are and whether they are surmountable.
> > 
> > 
> > Did we have this conversation? @jhesketh @jecluis @mgfritch ?
> 
> We did, and we've postponed it to a later date when at least one of us is annoyed enough to pursue this avenue. I think it's okay to close the PR though, given it's full of conflicts by now and will need to be redone anyway should we move forward with black at some point.

@jhesketh I leave the closing up to you

jhesketh:
> > > > Yeah, let's talk about this on our next meeting, and figure out what the concerns are and whether they are surmountable.
> > > 
> > > 
> > > Did we have this conversation? @jhesketh @jecluis @mgfritch ?
> > 
> > 
> > We did, and we've postponed it to a later date when at least one of us is annoyed enough to pursue this avenue. I think it's okay to close the PR though, given it's full of conflicts by now and will need to be redone anyway should we move forward with black at some point.
> 
> @jhesketh I leave the closing up to you

So the discussion has continued (in slack) and it was more or less decided that we'd use black with double quotes (which is the default).

I've fixed this PR and after enough approvals we'll consider this agreed upon and merge it.

src/pyproject.toml
jecluis: Can we, please? If no one particularly opposes to it, that is. I can live with the 88 columns, but it's just easier to  have the 3 pane splits which wrap at 80 each, instead of having to get rid of one because of annoying line wrapping. :)

jecluis (dismissed):
I'm going to approve this because we need it IMO, but I'd very much like the line length to be at 80 for, admittedly, my benefit. I know this is no longer 1985, or whatever, but screens still play nice in factors of 80 for some weird reason. :shrug:

votdev:
Please merge after #578.

jhesketh:
Reset line length to 80chars, and reran black after 578 has merged as requested.

jecluis:
jenkins run tumbleweed

jecluis:
Let's prioritize merging this one PR over further gravel-related PRs to avoid the pains of rebasing it and avoid potential tox issues.

jhesketh:
jenkins test tumbleweed

jhesketh:
Fixed conflict. Sorry, didn't notice it til now.

jecluis:
jenkins test tumbleweed

jhesketh:
jenkins/setup-vagrant-tumbleweed appears to be failing at the same place for multiple PR's.

The root cause does need diagnosing, but given the annoyance of refactoring this large change we could consider using our super privs to merge this.

jecluis:
> jenkins/setup-vagrant-tumbleweed appears to be failing at the same place for multiple PR's.
> 
> The root cause does need diagnosing, but given the annoyance of refactoring this large change we could consider using our super privs to merge this.

I agree. Let's deal with the fallout afterwards.

jecluis (approved):
let's do this.

PR #509, opened 2021-05-22
glass: show devices step before deploying cluster

jecluis:
It makes more sense to show the devices that are going to be consumed *before* we deploy the cluster.

This patchset does said shift, and makes the backend assimilate the devices automatically during deployment.

We are also getting rid of mentions to `bootstrapping`, and preferring `deploying` instead/


src/glass/src/app/pages/install-wizard-page/install-wizard-page.component.ts
```diff
@@ -83,7 +80,7 @@ export class InstallWizardPageComponent implements OnInit {
             this.context.stage = 'bootstrapped';
             this.context.stepperVisible = true;
             // Jump to the 'Devices' step.
```
votdev, line 82:
```suggestion
            // Jump to the 'Services' step.
```

src/glass/src/app/pages/install-wizard-page/install-wizard-page.component.ts
```diff
@@ -102,19 +99,19 @@ export class InstallWizardPageComponent implements OnInit {
...
                     )
                   )
                 );
                 this.pollBootstrapStatus();
                 break;
-              case BootstrapStageEnum.done:
+              case NodeStageEnum.deployed:
                 this.context.stage = 'bootstrapped';
                 this.context.stepperVisible = true;
                 // Jump to the 'Devices' step.
```
votdev, line 111:
```suggestion
                // Jump to the 'Services' step.
```

votdev (changes requested):
Please run `npm run fix` to fix indention and other issues that can be done automatically.

src/glass/src/app/shared/services/local-inventory.service.ts
```diff
@@ -0,0 +1,115 @@
+/*
```
votdev, line 1: Please relocate this file to `src/app/shared/services/api`.

src/glass/src/app/shared/services/local-inventory.service.ts
```diff
@@ -0,0 +1,115 @@
...
+          console.log('checking node status: ', status);
+          if (status.inited) {
+            this.obtainInventory();
+          }
+        }
+      });
+  }
+
+  getDevices(): Observable<Volume[]> {
+    return new Observable<Volume[]>( (observer) => {
```
votdev: Please use `rxjs` to simplify the code.
```
return this.subject.pipe(
      map((inventory?: Inventory) => {
        if (!inventory) {
          return [];
        }
        return inventory.disks;
      })
    );
```
You need to add `import { map } from 'rxjs/operators';`, too.

src/glass/src/app/shared/services/local-inventory.service.ts
```diff
@@ -0,0 +1,115 @@
...
+  name: string;
+  info: Nic;
+};
+
+@Injectable({
+  providedIn: 'root'
+})
+export class LocalInventoryService {
+
+  private subject: BehaviorSubject<Inventory|undefined> =
```
votdev: More see in the `getInventory` method comment.
```suggestion
  private inventorySource: BehaviorSubject<Inventory|undefined> =
```

src/glass/src/app/shared/services/local-inventory.service.ts
```diff
@@ -0,0 +1,115 @@
...
+            });
+          });
+          observer.next(entries);
+          subscription.unsubscribe();
+        }
+      });
+    });
+  }
+
+  getInventory(): Observable<Inventory> {
```
votdev, line 83: Replace this whole thing with this at the top of the class.
```suggestion
private inventorySource: BehaviorSubject<Inventory | undefined> = new BehaviorSubject<Inventory | undefined>(undefined);
public inventory$: Observable<Inventory>;

constructor(...) {
   ...
   this.inventory$ = this.inventorySource.asObservable.
}
```

src/glass/src/app/shared/services/local-inventory.service.ts
```diff
@@ -0,0 +1,115 @@
...
+            return;
+          }
+          observer.next(inventory.disks);
+          subscription.unsubscribe();
+        }
+      });
+    });
+  }
+
+  getNICs(): Observable<NICEntry[]> {
```
votdev, line 62: Replace this by the following code after all other comments have been applied.
```suggestion
    this.inventory$.pipe(
      map((inventory: Inventory) => {
        const entries: NICEntry[] = [];
        Object.keys(inventory.nics).forEach((key: string) => {
          entries.push({
            name: key,
            info: inventory.nics[key]
          });
        });
        return entries;
      })
    );
```

votdev (changes requested):

src/glass/src/app/shared/services/local-inventory.service.ts
```diff
@@ -0,0 +1,104 @@
...
+          console.log('checking node status: ', status);
+          if (status.inited) {
+            this.obtainInventory();
+          }
+        }
+      });
+  }
+
+  getDevices(): Observable<Volume[]> {
+    return new Observable<Volume[]>((observer) => {
```
votdev, line 49:
```suggestion
    return this.subject.pipe(
      map((inventory?: Inventory) => {
        if (!inventory) {
          return [];
        }
        return inventory.disks;
      })
    );
```

jecluis:
@votdev thanks for your patience. Mind another look? Comments have been addressed.

src/glass/src/app/shared/services/api/local-inventory.service.ts
```diff
@@ -0,0 +1,95 @@
...
+            this.obtainInventory();
+          }
+        }
+      });
+
+    this.inventory$ = this.inventorySource.asObservable();
+  }
+
+  getDevices(): Observable<Volume[]> {
+    return this.inventorySource.pipe(
```
votdev, line 60: Sorry, my mistake, please use this:
```suggestion
    return this.inventory$.pipe(
```

src/glass/src/app/shared/services/api/local-inventory.service.ts
```diff
@@ -0,0 +1,95 @@
...
+          if (status.inited) {
+            this.obtainInventory();
+          }
+        }
+      });
+
+    this.inventory$ = this.inventorySource.asObservable();
+  }
+
+  getDevices(): Observable<Volume[]> {
```
votdev, line 59: I'm currently not sure how to progress here. If we stay on using this implementation we should add a documentation for this method which tells the developers that they subscribe to a subject. This is important because the subscription will be called automatically whenever the subject has been changed. This behavior can cause unexpected side-effects if this is not documented.

Otherwise we should use `ReplaySubject` which lead to the following implementation:
```
diff --git a/src/glass/src/app/shared/services/api/local-inventory.service.ts b/src/glass/src/app/shared/services/api/local-inventory.service.ts
--- a/src/glass/src/app/shared/services/api/local-inventory.service.ts	(revision a14335c9f96c8053c4bd43f9ee3ebd4320028ebc)
+++ b/src/glass/src/app/shared/services/api/local-inventory.service.ts	(date 1621954017930)
@@ -13,7 +13,7 @@
  * GNU General Public License for more details.
  */
 import { Injectable } from '@angular/core';
-import { BehaviorSubject, Observable } from 'rxjs';
+import { Observable, ReplaySubject } from 'rxjs';
 import { map } from 'rxjs/operators';
 
 import {
@@ -34,11 +34,9 @@
   providedIn: 'root'
 })
 export class LocalInventoryService {
-  public inventory$: Observable<Inventory | undefined>;
+  public inventory$: Observable<Inventory>;
 
-  private inventorySource: BehaviorSubject<Inventory | undefined> = new BehaviorSubject<
-    Inventory | undefined
-  >(undefined);
+  private inventorySource: ReplaySubject<Inventory> = new ReplaySubject<Inventory>(1);
 
   constructor(private localService: LocalNodeService, private pollService: PollService) {
     this.localService
@@ -57,22 +55,14 @@
   }
 
   getDevices(): Observable<Volume[]> {
-    return this.inventorySource.pipe(
-      map((inventory?: Inventory) => {
-        if (!inventory) {
-          return [];
-        }
-        return inventory.disks;
-      })
+    return this.inventory$.pipe(
+      map((inventory: Inventory) => inventory.disks)
     );
   }
 
   getNICs(): Observable<NICEntry[]> {
     return this.inventory$.pipe(
-      map((inventory?: Inventory) => {
-        if (!inventory) {
-          return [];
-        }
+      map((inventory: Inventory) => {
         const entries: NICEntry[] = [];
         Object.keys(inventory.nics).forEach((key: string) => {
           entries.push({
```

I think using `ReplaySubject` is the better approach because it behaves that way i expect when i subscribe to a method like `getDevices` and `getNICS`; they respond when the information/data is available.

votdev (changes requested):

src/glass/src/app/shared/services/api/local-inventory.service.ts
votdev: Some more thoughts. Is it intended to fetch the disks and nics only once and return that cached data on every call of `getDevices` and `getNICs`? If so, this is another + for using `ReplaySubject`.

jecluis (commented):
asd

jecluis:
@votdev updated

mgfritch (approved):
> It makes more sense to show the devices that are going to be consumed _before_ we deploy the cluster.
> 
will be interesting later when we handle this during a node join ;)


> We are also getting rid of mentions to `bootstrapping`, and preferring `deploying` instead/
> 
nice, I like the change in terminology :+1:

votdev (approved):

PR #510, opened 2021-05-24
images/builder: Build an image in one go with containers

jecluis:
This patchset introduces a container image able to automatically build a source tree upon being run.

Additionally, allows the user to configure the source repository's root directory and the builds' output directory via the CLI to the existing `build-image.sh` script.

This is mighty useful to build images in one go without having to 1) do a lot of things, and 2) require root. Also allows for better automation.

### DIFF
--- a/images/builder/Dockerfile
+++ b/images/builder/Dockerfile
@@ -1,0 +1,34 @@
+# project aquarium's container builder
+# Copyright (C) 2021 SUSE, LLC.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+FROM registry.opensuse.org/opensuse/leap:15.3
+LABEL aquarium.image.author="Joao Eduardo Luis <joao@suse.com>"
+LABEL aquarium.url="https://aquarist-labs.io"
+LABEL aquarium.github="https://github.com/aquarist-labs/aquarium"
+LABEL version="0.1.0"
+LABEL description="Aquarium image builder"
+
+
+RUN zypper ref
+RUN zypper --non-interactive install \
+  git python3 python3-pip python3-rados \
+  python3-kiwi nodejs-common npm \
+  sudo which gptfdisk kpartx dosfstools e2fsprogs \
+  make qemu-tools xfsprogs
+
+
+RUN mkdir -p /builder/{cache,bin,src,out}
+VOLUME [ "/builder/cache", "/builder/bin", "/builder/src", "/builder/out" ]
+
+ENTRYPOINT [ "/builder/bin/autobuilder.sh" ]
+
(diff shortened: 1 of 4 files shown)


images/builder/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+FROM registry.opensuse.org/opensuse/leap:15.3
+LABEL aquarium.image.author="Joao Eduardo Luis <joao@suse.com>"
+LABEL aquarium.url="https://aquarist-labs.io"
+LABEL aquariumg.github="https://github.com/aquarist-labs/aquarium"
```
mgfritch, line 17:
```suggestion
LABEL aquarium.github="https://github.com/aquarist-labs/aquarium"
```

images/builder/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+LABEL version="0.1.0"
```
mgfritch, line 18:
```suggestion
LABEL version=$(git describe --match 'v*')
```
keep in sync with the aquarium version?

tools/build-image.sh
```diff
@@ -27,7 +27,10 @@ usage: $0 [options]
 options:
   -n NAME | --name NAME     Specify build name (default: aquarium).
   -c | --clean              Cleanup an existing build directory before building.
-  -t | --type IMGTYPE       Specify image type (default: vagrant)
+  -t | --type IMGTYPE       Specify image type (default: vagrant).
```
mgfritch, line 30: add a `container` type  ?? :grin:

mgfritch (approved):
neat!

images/builder/Dockerfile
jecluis: this isn't valid syntax, apparently :( Probably needs to go through an ARG or something, and I'm not sure it's worth adding a new argument to the build to make this work.

tools/build-image.sh
jecluis: we're not actually building a container with aquarium in it, so it feels weird adding this to the build script (which is actually used to build aquarium by the container's entrypoint). IMO, this might need a script that calls either on the container builder or on `build-image.sh` though.

jhesketh (changes requested):
Looks good to me, but I think we can reduce some confusion here.

Is there any need to support building in any way that is /not/ in a container? If so, then we should move the current `build-image.sh` into `images/builder/`, merging it with autobuilder (I don't see why we need both).

Then supply a new `builder.sh` that:
1) Builds the podman container image (if required)
2) Takes in the outdir, srcdir etc..
3) Performs the build with podman

This would in effect make podman a requirement for building aquarium (but I'm fine with that). But it should also simplify our kiwi requirements and possibly our cross distro development support.

jecluis:
Those are fair points, but I find that they are beyond the scope of this PR because the proposal brings about issues it does not intend to solve right now, namely permission issues.

This PR requires podman to be run in privileged mode, which means all the files created by it will be owner by root. While the problem is not necessarily having root superpowers, which we already require due to kiwi, we do end up with things like glass's `node_modules/` and `dist/` being owned by root, which can be incredibly annoying during development.

I do think that is a good next incremental step though.

jhesketh:
Fair enough, lets do them as a follow up then.

jhesketh (approved):

Issue #404, opened 2021-04-11
gravel: nfs-ganesha pool is created with size 3 by default


jecluis:
The result is that we end up with an health warning once we add a second node, because the replication domain moves from `osd` to `host`, and we'll end up having this warning until we add a third node. This makes little sense to me.

The reason is simple: all pools are, by default and unless something is provided stating otherwise, created with size 3 by default.

This is configurable via the config option `osd_pool_default_size`.

Now, we're at a point where we must decide what we want to do about this: leave it as is, or adjust as we go?

My take is that we should, for a single node, drop it to 2, and increase it to 3 when we add a second node. However, this means we must be aware of which pools are not created by the user so we can adjust the redundancy ourselves on those. I think we already have all the infrastructure needed for this, by, essentially, running through all the pools and checking whether they match a service; if not, adjust.

Other thoughts?

@l-mb you might have a strong opinion on this one?

l-mb:
For two nodes, size=3 can also not be reliably satisfied. That only starts making sense with at least three nodes?

jecluis:
That's right. Push it to 3 on adding a third node makes more sense.

PR #511, opened 2021-05-24
gravel: osd pool default size

mgfritch:
set the osd pool size to 3 when more than 3 nodes are present,
otherwise, (re)set the pool size to 2.

- sets the `osd_pool_default_size` for newly created pools
- adjusts the pool size for any other existing pool(s):
  - if not managed by an aquarium service
  - if not a user-defined pool

Resolves: #404

## Checklist

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>

### DIFF
--- a/src/gravel/controllers/nodes/mgr.py
+++ b/src/gravel/controllers/nodes/mgr.py
@@ -18,6 +18,7 @@ from uuid import UUID, uuid4
 import random
 from typing import (
     Dict,
+    List,
     Optional,
 )
 from pathlib import Path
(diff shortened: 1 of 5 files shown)


src/gravel/controllers/nodes/mgr.py
```diff
@@ -650,3 +654,49 @@ async def _handle_ready_to_add(
...
+        """ reset the osd pool default size """
+        assert self._init_stage == NodeInitStage.STARTED
+
+        def get_target_size():
+            orch: Orchestrator = Orchestrator(self.gstate.ceph_mgr)
+            orch_hosts: List[OrchHostListModel] = orch.host_ls()
+            return 2 if len(orch_hosts) < 3 else 3
+
+        mon: Mon = self.gstate.ceph_mon
+        current_size: int = mon.get_pool_default_size()
```
jecluis, line 670: This can technically be `None` in case of error, so it might be better to check it, or at least `assert` if so. Better yet would be to catch a raised exception.

jecluis (commented):
Looks generally good but I think it's worth manually testing this before giving the a-okay.

src/gravel/controllers/nodes/mgr.py
mgfritch: good catch, I've added handling similar to how we deal failures during `set_replicated_ruleset` et al.

jecluis (approved):

PR #518, opened 2021-05-29
gravel: create system disk on initial deployment

jecluis:
## note before:

This PR does not propose itself to address all the needs for a system disk, nor the full impact of having, creating, or maintaining one. Instead, this is the building block on which further work should be performed.

Consider this PR as a draft, open for comments.

Consider tests to be an on-going task.

## what we do

* generate a potential solution for disk usage from the available devices, including a candidate device for system disk.
* create a system disk on initial node deploy.
* mount aquarium-specific directories from the system disk to the running system's root tree, either by overlaying them or bind mounting them.

## what we don't do

* mount the system disk on boot, should it be present.
* do all of the above for a node joining the system.
* present the user with the generated solution, and allow them to have an opinion.

### DIFF
--- a/src/gravel/controllers/nodes/disks.py
+++ b/src/gravel/controllers/nodes/disks.py
@@ -1,0 +1,151 @@
+# project aquarium's backend
+# Copyright (C) 2021 SUSE, LLC.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+from enum import Enum
+from logging import Logger
+from typing import List, Optional, Tuple
+from fastapi.logger import logger as fastapi_logger
+from pydantic import BaseModel, Field
+
+from gravel.cephadm.models import NodeInfoModel, VolumeDeviceModel
+from gravel.controllers.gstate import GlobalState
+
+
+logger: Logger = fastapi_logger
+
+
+class DiskTypeEnum(int, Enum):
+    NONE = 0
+    HDD = 1
+    SSD = 2
+
+
+class DiskInfoModel(BaseModel):
+    vendor: str = Field("", title="disk's manufacturer")
+    model: str = Field("", title="disk's model")
+
+
+class DiskModel(BaseModel):
+    path: Optional[str] = Field(None, title="disk's device path")
+    size: int = Field(0, title="size in bytes")
+    type: DiskTypeEnum = Field(DiskTypeEnum.NONE, title="disk type")
+    info: DiskInfoModel = Field(title="additional info about disk")
+
+
+class RejectedDiskModel(BaseModel):
+    disk: DiskModel = Field(title="rejected disk")
+    reasons: List[str] = Field([], title="reasons for rejection")
+
+
+class DiskSolution(BaseModel):
+    systemdisk: Optional[DiskModel] = Field(None, title="chosen system disk")
+    storage: List[DiskModel] = Field([], title="chosen storage disks")
+    storage_size: int = Field(0, title="total size for storage, in bytes")
+    rejected: List[RejectedDiskModel] = Field([], title="rejected disks")
+    possible: bool = Field(False, title="deployment possible")
+
+
+def _device_to_disk(device: VolumeDeviceModel) -> DiskModel:
+
+    def _get_disk_type(rotational: bool) -> DiskTypeEnum:
+        return DiskTypeEnum.HDD if rotational else DiskTypeEnum.SSD
+
+    return DiskModel(
+        path=device.path,
+        size=device.sys_api.size,
+        type=_get_disk_type(device.sys_api.rotational),
+        info=DiskInfoModel(
+            vendor=device.sys_api.vendor,
+            model=device.sys_api.model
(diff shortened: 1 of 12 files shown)


src/gravel/controllers/nodes/disks.py
```diff
@@ -0,0 +1,135 @@
...
+        size=device.sys_api.size,
+        type=_get_disk_type(device.sys_api.rotational),
+        info=DiskInfoModel(
+            vendor=device.sys_api.vendor,
+            model=device.sys_api.model
+        )
+    )
+
+
+class Disks:
```
jhesketh, line 74: (nit) Why does this need to be a class? Since we're not creating any instances wouldn't it make more sense to just have gen_solution as a standard method?

src/gravel/controllers/nodes/systemdisk.py
```diff
@@ -0,0 +1,255 @@
...
+            # create lvm volume
+            await self.lvm(f"pvcreate {devpath}")
+            await self.lvm(f"vgcreate aquarium {devpath} --addtag @aquarium")
+            await self.lvm("lvcreate -l 100%FREE -n systemdisk aquarium")
+
+            lvmdev: Path = Path("/dev/mapper/aquarium-systemdisk")
+            assert lvmdev.exists()
+
+            # format with xfs
+            await aqr_run_cmd(shlex.split(f"mkfs.xfs -m bigtime=1 {lvmdev}"))
```
jhesketh, line 152: (nit) this is probably simple enough for our environment, but we could also look at using something like pyparted. (Not sure about LVM support though)

src/gravel/controllers/nodes/systemdisk.py
```diff
@@ -0,0 +1,255 @@
...
+                continue
+            lst.append(MountEntry(source=fields[0], dest=fields[1]))
+    return lst
+
+
+class SystemDisk:
+
+    _gstate: GlobalState
+    _overlaydirs: Dict[str, str] = {
+        "etc": "/etc",
```
jhesketh, line 84: (nit) Could just use a list and then `os.path.basenam()`

src/gravel/controllers/nodes/systemdisk.py
```diff
@@ -0,0 +1,255 @@
...
+            raise MountError(msg="/aquarium mount point does not exist")
+
+        try:
+            await aqr_run_cmd(shlex.split(f"umount {aqrmnt}"))
+        except Exception as e:
+            raise MountError(msg=str(e))
+
+    # lacking a better name atm, we're calling this function 'swap in', as in
+    # we're swapping these directories in, some being overlayed, others bind
+    # mounted.
```
jhesketh, line 202: Do we need a `swapout` as well?

(nit) An alternative could be `enable`?

jhesketh (approved):
I haven't tested this but in general it looks good.

(Unless I'm mistaken,) The disadvantage of this approach is that it uses an entire disk for aquarium data in addition to wherever aquarium itself is installed.

Would it make sense to explore options where our appliance disk image includes a partition for data?

jecluis:
> I haven't tested this but in general it looks good.
> 
> (Unless I'm mistaken,) The disadvantage of this approach is that it uses an entire disk for aquarium data in addition to wherever aquarium itself is installed.
> 
> Would it make sense to explore options where our appliance disk image includes a partition for data?

Aquarium is not installed, nor will it be running from a system partition. Instead, we're running it from a ramdisk, either supplied via a usb stick or whatever. So there's no additional disk. If there was, we would not be needing to do this, and we'd just do this stuff from said disk (or partition the disk and keep this stuff in a separate partition).

jhesketh:
> > I haven't tested this but in general it looks good.
> > (Unless I'm mistaken,) The disadvantage of this approach is that it uses an entire disk for aquarium data in addition to wherever aquarium itself is installed.
> > Would it make sense to explore options where our appliance disk image includes a partition for data?
> 
> Aquarium is not installed, nor will it be running from a system partition. Instead, we're running it from a ramdisk, either supplied via a usb stick or whatever. So there's no additional disk. If there was, we would not be needing to do this, and we'd just do this stuff from said disk (or partition the disk and keep this stuff in a separate partition).

Ah okay, that's my misunderstanding. I was working on the assumption that an option for aquarium would be to install the image to disk, kinda like the way VM appliances work.

src/gravel/controllers/nodes/systemdisk.py
jecluis: Yeah, that's true, but I found this gives more visibility into what's happening, and the impact is negligible.

src/gravel/controllers/nodes/systemdisk.py
jecluis: This was the only way I found to be able to run lvm that isn't N-years old without evidence of being maintained. I'll look into pyparted, but given it seems to be solely for libparted I'm inclined with going with the current approach.

src/gravel/controllers/nodes/systemdisk.py
jecluis: yeah, sounds like a good alternative.

src/gravel/controllers/nodes/disks.py
jecluis: Mostly future looking, that's all.

tserong (commented):
I haven't tested, but LGTM in general (one immediate thought is that it's just taking the smallest disk, where we'd probably want to preference SSD over HDD...?)

jecluis:
> I haven't tested, but LGTM in general (one immediate thought is that it's just taking the smallest disk, where we'd probably want to preference SSD over HDD...?)

yeah, that's an oversight. I'll try reworking that.

tserong (approved):
LGTM. Further refinements can happen in subsequent PRs.

PR #521, opened 2021-05-31
image: move image from 'microOS/' to 'aquarium'

jecluis:
Also drops references to MicroOS in the image, and updates the README.


images/README.md
```diff
@@ -1,21 +1,35 @@
...
+needed for Aquarium's execution, so we have them available upon first run,
+including an image for `Ceph` and an image for `etcd`.
+
+
+## Containerized build environment
+
+Under `builder/` we provide a `Dockerfile` that will be able to automatically
+build an image from a checked out branch. This might be useful for those on
+distributions that do not support `kiwi`, or under which `kiwi` does not work
+reliably.
```
jhesketh, line 35: My opinion is to make this the only way we build images so that we aren't maintaining multiple environments. (Can be a separate change though).

images/README.md
```diff
@@ -1,21 +1,35 @@
...
-of writing, simply installing additional packages that are not available through
-the repositories, relying on `pip` to do so. In the future, we expect to use
-this file to enable the backend's service.
-
-One can also find a `build.sh` script, which is merely a helper to build a
-microOS image from `config.xml`. At the moment, a Vagrant image will be
-generated. Feel free to run `build.sh --help` for slightly more info, but the
-script itself should be fairly straightforward.
+To help us build an image reliably, we are relying on `kiwi`. We do so by
+offering an image to natively build using `kiwi`, and containerized
```
jhesketh, line 4: This sentence is confusing to me. We don't offer an image to build with kiwi, rather we supply a way to build with kiwi on the developers machine.

jhesketh (commented):
I'm not sure if we should reject this change in favour of simply making building inside a container the default and only supported way (as previously commented on the other PR).

jecluis:
> I'm not sure if we should reject this change in favour of simply making building inside a container the default and only supported way (as previously commented on the other PR).

As mentioned in the other PR, building solely in a container brings about permission issues I do not have a way to solve right now, nor the bandwidth to do so.

And I really need to be building images during my development without being too concerned that permissions are being changed under my feet. The builder is great for automating builds, but atm not great to build while developing.

images/README.md
jecluis: I'll address it.

jhesketh:
Would it be sufficient to have a build script that calls podman and then chmod's the files back to something reasonable?

jecluis:
Maybe, but I find that beyond what we're doing here. That's the kind of thing that should be its own PR.

jecluis:
> # [Codecov](https://codecov.io/gh/aquarist-labs/aquarium/pull/521?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=aquarist-labs) Report
> 
> > Merging [#521](https://codecov.io/gh/aquarist-labs/aquarium/pull/521?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=aquarist-labs) ([0a74d8c](https://github.com/aquarist-labs/aquarium/commit/0a74d8c537e3801e525004339c0760b4aa112074)) into [main](https://codecov.io/gh/aquarist-labs/aquarium/commit/3ccd56205c7fceadbd36262d75e8fd0e3f65f0ea?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=aquarist-labs) ([3ccd562](https://github.com/aquarist-labs/aquarium/commit/3ccd56205c7fceadbd36262d75e8fd0e3f65f0ea)) will **decrease** coverage by `0.22%`.

This report must be borked. There's no way this is even touching close to gravel.

images/aquarium/config.xml
```diff
@@ -43,7 +43,7 @@
         </profile>
     </profiles>
     <preferences>
-        <version>16.0.0</version>
+        <version>0.1.0</version>
```
tserong, line 46: This should ideally come from the latest version git tag, probably by renaming this file to images/aquarium/config.xml.in, changing this line to \<version\>VERSION\</version\> (or whatever), then adding an extra bit of script somewhere to s/VERSION/$(git describe [...])/, but I'm fine if this is in a separate PR.

tserong (approved):

tserong:
Just for the record, I didn't test this myself (sorry, should have mentioned that in the approval)

images/README.md
tserong: Agreed on both counts

PR #513, opened 2021-05-25
ceph.git: update to the latest ceph pacific

mgfritch:
- Ceph Pacific v16.2.4 + latest backports
- includes upstream backport of:
  - ceph/ceph#40555
  - ceph/ceph#40463

Resolves: #389
Signed-off-by: Michael Fritch <mfritch@suse.com>

## Checklist
- [ ] References issues, create if required
- [ ] Updates documentation if necessary
- [ ] Includes tests for new functionality or reproducer for bug

---

<details>
<summary>Show available Jenkins commands</summary>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>

PR #526, opened 2021-06-01
gravel: cephadm/bootstrap: parser expects line before EOF

jecluis:
The config parser expects a blank line before EOF, otherwise it will
fail to parse the user-supplied config.

This is required as a prerequisite to merge #513 

Signed-off-by: Joao Eduardo Luis <joao@suse.com>

### DIFF
--- a/src/gravel/cephadm/cephadm.py
+++ b/src/gravel/cephadm/cephadm.py
@@ -131,6 +131,7 @@ class Cephadm:
             s = (
                 "[global]",
                 "osd_pool_default_size = 2",
+                ""
             )
             return "\n".join(s)
 


jhesketh (approved):
Seems like an upstream bug also.

jecluis:
how's that?

jhesketh:
> how's that?

cephadm shouldn't fail if there's a missing line before EOF. It should still be parsable.

jecluis:
> > how's that?
> 
> cephadm shouldn't fail if there's a missing line before EOF. It should still be parsable.

afaict, technically it just needs a new-line at end of file. I'm guessing because somewhere the config is being parsed with the equivalent of a `splitlines()`. That doesn't sound like a bug to me, but expected behavior :shrug:

jhesketh:
> > > how's that?
> > 
> > 
> > cephadm shouldn't fail if there's a missing line before EOF. It should still be parsable.
> 
> afaict, technically it just needs a new-line at end of file. I'm guessing because somewhere the config is being parsed with the equivalent of a `splitlines()`. That doesn't sound like a bug to me, but expected behavior shrug

Right, but it's a papercut for users.. Why should a user be expected to add a blank line when cephadm can easily just allow for it either way.

(We should carry this fix regardless of course)

jecluis:
> > > > how's that?
> > > 
> > > 
> > > cephadm shouldn't fail if there's a missing line before EOF. It should still be parsable.
> > 
> > 
> > afaict, technically it just needs a new-line at end of file. I'm guessing because somewhere the config is being parsed with the equivalent of a `splitlines()`. That doesn't sound like a bug to me, but expected behavior shrug
> 
> Right, but it's a papercut for users.. Why should a user be expected to add a blank line when cephadm can easily just allow for it either way.

Fair enough. Maybe @mgfritch can look into that upstream?

p-se (approved):

PR #547, opened 2021-06-07
tools/libaqr: check deployment start error

kshtsk:
If vagrant up fails when trying to start a deployment status is not
checked. Thif patch address this issue.

## Checklist

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>

### DIFF
--- a/tools/libaqr/deployment.py
+++ b/tools/libaqr/deployment.py
@@ -147,7 +147,9 @@ class Deployment:
             with vagrant.deployment(self._path) as deployment:
                 if deployment.running or deployment.preparing:
                     raise DeploymentRunningError()
-                deployment.start(conservative=conservative)
+                ret, err = deployment.start(conservative=conservative)
+                if not ret:
+                    raise AqrError('Cannot start deployment: unknown vagrant error')
         except AqrError as e:
             raise e
 


kshtsk:
jenkins test tumbleweed

tools/libaqr/deployment.py
```diff
@@ -147,7 +147,9 @@ def start(self, conservative: bool) -> None:
             with vagrant.deployment(self._path) as deployment:
                 if deployment.running or deployment.preparing:
                     raise DeploymentRunningError()
-                deployment.start(conservative=conservative)
+                ret, err = deployment.start(conservative=conservative)
+                if not ret:
+                    raise AqrError('Cannot start deployment: unknown vagrant error')
```
jecluis, line 152: is the error not returned via `err`?

kshtsk: what if it is empty?

jecluis: then you can say it's unknown, but if it's not, then it is not unknown.

kshtsk: You're forcing interactive mode:
```python

    def start(self, conservative: bool) -> Tuple[bool, Optional[str]]:

        cmd = "vagrant up"
        if conservative:
            cmd += " --no-parallel --no-destroy-on-error"

        retcode, _, err = self._run(cmd, interactive=True)
        if retcode != 0:
            return False, err
        return True, None
 ```

Comment (jecluis) [reply tools/libaqr/deployment.py:152]: Where are you getting at? Does that return an empty string every time?

Comment (kshtsk) [reply tools/libaqr/deployment.py:152]: which means you do not care what the stderr is, I don't know why do you want it, so it is always empty.

Comment (kshtsk) [reply tools/libaqr/deployment.py:152]:
```python
    def _run(self, cmd: str, interactive: bool = False) -> Tuple[int, str, str]:
        with self.safeenv() as env:
            capture: Optional[int] = subprocess.PIPE
            if interactive:
                capture = None
            proc = subprocess.run(
                shlex.split(cmd),
                stderr=capture,
                stdout=capture,
                env=env
            )
            stdout = "" if interactive else proc.stdout.decode("utf-8")
            stderr = "" if interactive else proc.stderr.decode("utf-8")
            return proc.returncode, stdout, stderr
```
I am getting it from the source code of libaqr/vagrant.py

jecluis: got it

jecluis (approved):

kshtsk:
jenkins test tumbleweed

kshtsk:
jenkins run tumbleweed

PR #548, opened 2021-06-07
gravel: handle systemdisk on boot

jecluis:
This patchset makes the image recognize the system disk on boot, and overlays or bind mounts the right directories inn the right places.

### DIFF
--- a/doc/dev/gravel/systemd-boot.md
+++ b/doc/dev/gravel/systemd-boot.md
@@ -1,0 +1,45 @@
+# Aquarium systemd boot process
+
+## aquarium-boot
+
+`systemd` unit that will be running after `local-fs.target`, runs
+`/usr/share/aquarium/boot/aqrbootsetup.sh`, which tries to find an LVM disk
+tagged with `@aquarium`. If such a disk is found, the disk will be mounted in
+`/aquarium` and several operation critical directories will either be overlayed
+over existing system directories, or bind mounted to specific places.
+
+### Overlayed directories
+
+These are
+
+* `/etc`, so we can keep track of system configuration files that were changed
+  for Aquarium's operations. This includes Ceph's systemd unit files.
+
+* `/var/log`, so logs are persisted between runs.
+
+* `/var/lib/etcd`, where we keep this node's etcd persistent db.
+
+* `/var/lib/containers`, where container images are kept.
+
+### Bind mounted directories
+
+* `/var/lib/ceph`, where we'll keep Ceph's persistent data. This directory
+  cannot be overlayed, because overlaying causes Ceph to misbehave. Reasons
+  unknown.
+  
+### Known issues
+
+Unfortunately we have been unable to ensure we overlay `/etc` at a time before
+`systemd` reads the contents of `/etc/systemd`. Therefore, we are overlaying
+`/etc` contents after systemd already has a list of units to be started. This
+means that the Ceph units are not started automatically during boot, because
+from `systemd`'s point of view they did not exist when it looked for them.
+
+This means the onus of starting the `ceph.target` falls on Aquarium.
+
+
+## aquarium
+
+`systemd` unit that will be running after `network.target`, runs Aquarium from
+`/usr/lib/share/aquarium`.
+
--- a/src/boot/aqrbootsetup.sh
+++ b/src/boot/aqrbootsetup.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+#
+# project aquarium's boot setup
+# Copyright (C) 2021 SUSE, LLC.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+
+function overlay() {
+  lower=${1}
+  upper=${2}
+
+  aqrdir=/aquarium/${upper}
+  mount -t overlay \
+    -o lowerdir=${lower},upperdir=${aqrdir}/overlay,workdir=${aqrdir}/temp \
+    overlay ${lower} || return 1
+}
+
+systemdisk=$(lvm lvs --noheadings -o vg_name,lv_name @aquarium |
+             sed 's/^[ ]\+//' | tr ' ' '-')
+
+[[ -z "${systemdisk}" ]] && \
+  echo "Aquarium system disk not found, assuming new deployment" && \
+  exit 0
+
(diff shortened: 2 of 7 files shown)


jecluis:
jenkins run leap

AvengerMoJo:
Looking good. I can try to pull and test it as well. 
I have done something like that before for the container directory. I don't exactly the root cause but my estimation was it is using overlay itself and overlay on top making pod pulling process could be flicky. 

On the same note, I will add this to another issue. Now I'm thinking about it when the first Node Container is being pulled from the internet. The rest of the nodes can actually share and overlay the same shared directory. Which means the complete cluster will end up running the same pod with the same version using the same location but where is the question.

doc/dev/gravel/systemd-boot.md
```diff
@@ -0,0 +1,45 @@
+# Aquarium systemd boot process
+
+## aquarium-boot
+
+`systemd` unit that will be running after `local-fs.target`, runs
+`/usr/share/aquarium/boot/aqrbootsetup.sh`, which tries to find an lvm disk
```
votdev, line 6:
```suggestion
`/usr/share/aquarium/boot/aqrbootsetup.sh`, which tries to find a LVM disk
```

src/gravel/controllers/nodes/mgr.py
```diff
@@ -276,10 +278,17 @@ async def _node_start(self) -> None:
...
-        logger.info("start leader node")
+        logger.info("start connection manager")
         self._incoming_task = asyncio.create_task(self._incoming_msg_task())
         self._connmgr.start_receiving()
 
+    async def _start_ceph(self) -> None:
+        """ Start ceph's systemd target """
+        ret, _, err = await aqr_run_cmd(["systemctl", "start", "ceph.target"])
+        if ret:
+            raise NodeError(f"unable to start Ceph target: {err}")
```
votdev, line 289:
```suggestion
            raise NodeError(f"Unable to start Ceph target: {err}")
```

src/gravel/controllers/nodes/mgr.py
```diff
@@ -276,10 +278,17 @@ async def _node_start(self) -> None:
...
+        logger.info("started ceph target")
```
votdev, line 290:
```suggestion
        logger.info("Started ceph target")
```

src/gravel/controllers/nodes/mgr.py
```diff
@@ -276,10 +278,17 @@ async def _node_start(self) -> None:
 
         self._init_stage = NodeInitStage.STARTED
 
-        logger.info("start leader node")
+        logger.info("start connection manager")
```
votdev, line 281:
```suggestion
        logger.info("Start connection manager")
```

votdev (changes requested):
I hope i am not too pedantic with this change requests.

src/boot/aqrbootsetup.sh
```diff
@@ -0,0 +1,57 @@
...
+
+systemdisk=$(lvm lvs --noheadings -o vg_name,lv_name @aquarium |
+             sed 's/^[ ]\+//' | tr ' ' '-')
+
+[[ -z "${systemdisk}" ]] && \
+  echo "aquarium systemdisk not found, assuming new deployment" && \
+  exit 0
+
+[[ "${systemdisk}" != "aquarium-systemdisk" ]] && \
+  echo "unknown systemdisk, expected aquarium-systemdisk" >/dev/stderr && \
```
votdev, line 35:
```suggestion
  echo "Unknown systemdisk, expected aquarium-systemdisk" >/dev/stderr && \
```

src/boot/aqrbootsetup.sh
```diff
@@ -0,0 +1,57 @@
...
+  exit 1
+
+[[ ! -d "/aquarium" ]] && \
+  (mkdir /aquarium || (echo "error creating /aquarium" && exit 1))
+
+[[ ! -e "/dev/mapper/aquarium-systemdisk" ]] && \
+  echo "unable to find aquarium block device" && \
```
votdev, line 42:
```suggestion
  echo "Unable to find aquarium block device" && \
```

votdev (changes requested):

doc/dev/gravel/systemd-boot.md
jecluis: I'm fine with the "lvm" -> "LVM" change, but, and may @asettle correct me, it's still an "an LVM" on the account that it's read "an ell-v-m".

asettle: LVM :heavy_check_mark: 
an :heavy_check_mark:

AvengerMoJo (commented):
Looks good

doc/dev/gravel/systemd-boot.md
```diff
@@ -0,0 +1,45 @@
...
+
+* `/var/lib/etcd`, where we keep this node's etcd persistent db.
+
+* `/var/lib/containers`, where container images are kept.
+
+### Bind mounted directories
+
+* `/var/lib/ceph`, where we'll keep Ceph's persistent data. This directory
+  cannot be overlayed, because overlaying causes Ceph to misbehave. Reasons
+  unknown.
```
mgfritch, line 28: track via a github issue?

src/boot/aqrbootsetup.sh
```diff
@@ -0,0 +1,57 @@
...
+
+[[ ! -d "/aquarium" ]] && \
+  (mkdir /aquarium || (echo "error creating /aquarium" && exit 1))
+
+[[ ! -e "/dev/mapper/aquarium-systemdisk" ]] && \
+  echo "Unable to find aquarium block device" && \
+  exit 1
+
+# mount system disk
+mount /dev/mapper/aquarium-systemdisk /aquarium
```
mgfritch, line 46:
```suggestion
mount /dev/mapper/aquarium-systemdisk /var/lib/aquarium
```

LSB?

doc/dev/gravel/systemd-boot.md
```diff
@@ -0,0 +1,45 @@
...
+
+`systemd` unit that will be running after `local-fs.target`, runs
+`/usr/share/aquarium/boot/aqrbootsetup.sh`, which tries to find an LVM disk
+tagged with `@aquarium`. If such a disk is found, the disk will be mounted in
+`/aquarium` and several operation critical directories will either be overlayed
+over existing system directories, or bind mounted to specific places.
+
+### Overlayed directories
+
+These are
```
mgfritch, line 13:
```suggestion
```
(nit)

mgfritch (approved):
This will work, but would be nice to decouple aquarium from the other system components. I'm still a bit worried if the overlay will create additional side effects for ceph, systemd, ntp, http proxy, etc...

src/boot/aqrbootsetup.sh
jecluis: ugh, this would have been a good idea, but will likely need to be done in a next pass -- we'll need to change this on the creation code as well. I'll open an issue and address it when tackling the cluster extension bits.

Issue #450, opened 2021-04-25
gravel: don't use insecure global_id reclaim


jecluis:
With the fixing of [CVE-2021-20288](https://docs.ceph.com/en/latest/security/CVE-2021-20288/) , Ceph has fixed a vulnerability related to global_id reclaim.

With the fixing patch comes a neat warning about usage of insecure global_id reclaim. Why this is happening on newly deployed clusters is beyond me at this point.

```
[ceph: root@node3 /]# ceph -s
  cluster:
    id:     02cb9904-a57f-11eb-a5fb-5254000d565a
    health: HEALTH_WARN
            clients are using insecure global_id reclaim
            mons are allowing insecure global_id reclaim
```

How to fix: https://docs.ceph.com/en/latest/security/CVE-2021-20288/#fix-details

Basically we'll have to set `auth_allow_insecure_global_id_reclaim` to `false`. We need to check whether this is compatible with the `librados` version we are using (being consumed from tumbleweed's repositories).

tserong:
> Why this is happening on newly deployed clusters is beyond me at this point.

I assume it's because there's still no guarantee that all your potential new clients are up-to-date?

> Basically we'll have to set `auth_allow_insecure_global_id_reclaim` to `true`

Don't you mean `false`? Nathan did that in sesdev https://github.com/SUSE/sesdev/pull/596.

jecluis:
_sigh_ Yes I do. Updated the initial comment. Thanks! :)

jecluis:
Tumbleweed's librados doesn't seem to support the fixed global_id reclaim, so applying the patch for this ticket essentially breaks everything because librados is unable to reach the cluster.

tserong:
Is it not possible to check if the option exists, then only try to set it if it does? Or is that not worth the trouble, and we should just wait for Tumbleweed to be updated?

jecluis:
The option exists on any new pacific image, which we're using because we're still following v16. This automatically raises an error (which we could mute) until the option is set to `false`. Our problem is that because tumbleweed's librados is not fixed yet, we don't get to disable the unsafe global_id reclaim, and end up having the error all the time.

The alternative would be fixing the image prior to the latest v16 until we get the tumbleweed package.

tserong:
Ah! Now I get it (somehow I misunderstood your earlier comment, and hadn't quite connected everything together in my head). The new Tumbleweed package probably isn't _too_ far away (but I don't have an actual ETA).

[...]

PR #552, opened 2021-06-08
gravel: do not allow insecure global id in Ceph

jecluis:
Closes the hole of CVE-2021-20288.

Resolves: #450 


asettle (approved):

mgfritch (commented):
generally ok, but maybe we should set this via ceph.conf during bootstrap (rather than after)?

https://github.com/ceph/ceph/blob/a9ea94f8f6f61f5db8f40949e4d681dbd13a3a83/src/cephadm/cephadm#L3888-L3892

Issue #486, opened 2021-05-07
glass: Add capacity dashboard widget


votdev:
Add a capacity dashboard widget that displays the real usage (instead of the allocated).

PR #560, opened 2021-06-09
glass: Add capacity dashboard widget

votdev:
Bildschirmfoto vom 2021-06-09 14-35-05

- Rename existing 'Services Capacity' widget to 'Services Utilization'
- Add new 'Services Capacity' widget
- Hide legends in both widgets, display y-axis instead

Fixes: https://github.com/aquarist-labs/aquarium/issues/486

## Checklist

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>

### DIFF
--- a/src/glass/src/app/core/dashboard/dashboard.module.ts
+++ b/src/glass/src/app/core/dashboard/dashboard.module.ts
@@ -12,6 +12,7 @@ import { HostsDashboardWidgetComponent } from '~/app/core/dashboard/widgets/host
 import { PerformanceDashboardWidgetComponent } from '~/app/core/dashboard/widgets/performance-dashboard-widget/performance-dashboard-widget.component';
 import { ServicesCapacityDashboardWidgetComponent } from '~/app/core/dashboard/widgets/services-capacity-dashboard-widget/services-capacity-dashboard-widget.component';
 import { ServicesDashboardWidgetComponent } from '~/app/core/dashboard/widgets/services-dashboard-widget/services-dashboard-widget.component';
+import { ServicesUtilizationDashboardWidgetComponent } from '~/app/core/dashboard/widgets/services-utilization-dashboard-widget/services-utilization-dashboard-widget.component';
 import { SysInfoDashboardWidgetComponent } from '~/app/core/dashboard/widgets/sys-info-dashboard-widget/sys-info-dashboard-widget.component';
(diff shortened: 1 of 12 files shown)


jecluis (commented):
widgets look good to me.

Although I think this doesn't fully resolve #486 -- I think we need a capacity widget that represents the actual capacity being used in terms of cluster usage, in the shape of what we currently have to the "allocated storage" widget.

I don't think it needs to come in this PR, but something to keep in mind because merging this will close that issue.

votdev:
> widgets look good to me.
> 
> Although I think this doesn't fully resolve #486 -- I think we need a capacity widget that represents the actual capacity being used in terms of cluster usage, in the shape of what we currently have to the "allocated storage" widget.
> 
> I don't think it needs to come in this PR, but something to keep in mind because merging this will close that issue.

Fixed with second commit.

jecluis (approved):
Generally looks good to me. There seems to be a bug, either here or in the backend, that has us showing 6.99G as the total capacity (should have been the 23.99G), but that's something we can address at a later stage.

kshtsk:
jenkins run opensuse

votdev:
> Generally looks good to me. There seems to be a bug, either here or in the backend, that has us showing 6.99G as the total capacity (should have been the 23.99G), but that's something we can address at a later stage.

Maybe your expectation of the widget is different to mine, but shouldn't the `Capacity` widget show the total available capacity that can be used effectively?

The following setup looks fine based on the above assumption.

Bildschirmfoto vom 2021-06-10 10-49-36

Bildschirmfoto vom 2021-06-10 10-49-23

jecluis:
> > Generally looks good to me. There seems to be a bug, either here or in the backend, that has us showing 6.99G as the total capacity (should have been the 23.99G), but that's something we can address at a later stage.
> 
> Maybe your expectation of the widget is different to mine, but shouldn't the `Capacity` widget show the total available capacity that can be used effectively?

I think that's also a fair expectation. I do think however that this ends up being redundant information if we take into consideration the "Allocated Storage" widget plus the "Service Capacity". I think the user might want to know how much actual storage is being used in the cluster, for reals, which isn't provided by the allocated storage widget.

But I do see your point, and it makes sense. Maybe we just need to find a better way to convey these two different concepts to the user.

PR #570, opened 2021-06-15
gravel: setting hostname must update state

jecluis:
We need to update the in-memory state for the node setting a new
hostname; otherwise we will be joining or creating a new cluster with,
potentially, the old hostname.

Additionally, fix an annoying issue with the stepper not transitioning
when the join is successful.


src/gravel/api/orch.py
```diff
@@ -53,10 +53,6 @@ class HostsDevicesModel(BaseModel):
     devices: List[DeviceModel]
 
 
-class SetNtpRequest(BaseModel):
```
s0nea, line 62: Woops... thank you!

votdev (approved):

PR #574, opened 2021-06-16
aquarium: reflect reality, license under GPL3

jecluis:

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,8 +1,8 @@
 Name: aquarium
 
 Files: *
-Copyright: (c) 2021 SUSE, LLC.
-License: LGPL-2.1 or LGPL-3 (see COPYING-LGPL2.1 and COPYING-LGPL3)
+Copyright: (c) 2021 SUSE, LLC., unless otherwise specified
+License: GPL-3 (see COPYING-GPL3), unless otherwise specified
 
 Files: src/gravel/ceph.git
 Copyright: (c) Ceph and contributors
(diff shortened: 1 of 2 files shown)


l-mb:
Just for the record: have you checked that the new '*' coverage doesn't accidentally cover any files that are indeed on a different license?

asettle (dismissed):

jhesketh (dismissed):

jhesketh:
I wonder if this is considered something that needs every contributor to agree to in case it could have been interpreted as every file being dual license and we're removing one?

jecluis:
> I wonder if this is considered something that needs every contributor to agree to in case it could have been interpreted as every file being dual license and we're removing one?

@jhesketh So far all contributors' copyright holder has been SUSE, and the agreement when we were starting the project was GPL3; and pretty much all the files have headers under GPL3. I'd say this is a nonissue.

> Just for the record: have you checked that the new '*' coverage doesn't accidentally cover any files that are indeed on a different license?

@l-mb AFAICT we're not covering  anything else that is not covered by another portion of the COPYING file. But we can adjust the wording to `unless otherwise specified` just in case.

l-mb (approved):
LGTM, thanks for the clarification.

asettle:
@jecluis are you waiting on any further reviews or can this be merged?

jecluis:
@asettle @l-mb just repushed, adding an `unless otherwise specified` to the wildcard entry in the COPYING file.

asettle (approved):

PR #575, opened 2021-06-17
gravel: pass hostname as parameters to deploy & join (part 1)

jecluis:
We are now passing the user-specified hostname as parameters to the deploy and join endpoints, so we can set the hostname as part of a node's deployment process.

We are also removing the endpoint that allows to set the hostname because we suspect changing a hostname after a node has been deployed will not be a trivial task, and we currently have no other use for this endpoint.

~Depends on #581~

### DIFF
--- a/src/glass/src/app/pages/install-wizard/install-create-wizard-page/install-create-wizard-page.component.spec.ts
+++ b/src/glass/src/app/pages/install-wizard/install-create-wizard-page/install-create-wizard-page.component.spec.ts
@@ -67,15 +67,6 @@ describe('InstallCreateWizardPageComponent', () => {
   }));
 
   it('should show notification only once [3]', fakeAsync(() => {
-    component.startBootstrap();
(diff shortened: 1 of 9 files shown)


jhesketh (approved):

AvengerMoJo:
@jecluis do we need to care to sync the /etc/hosts? or possible DNS support in the future?

jecluis:
> @jecluis do we need to care to sync the /etc/hosts? or possible DNS support in the future?

I'm not sure. Maybe? Although this feels like something that would be external to the cluster, like the same with dhcp etc.

jecluis:
@votdev mind taking a look at the glass bits?

votdev (approved):

jecluis:
@votdev @s0nea addressed the cosmetic issues; mind reviewing/approving?

s0nea (approved):

Issue #421, opened 2021-04-14
glass: user management


votdev:
Create a page to manage (create/edit/delete) users.

https://github.com/aquarist-labs/aquarium/issues/23
https://github.com/aquarist-labs/aquarium/issues/26

Issue #26, opened 2021-02-01
gravel: user management


votdev:
A user management is required to be able to implement an UI authentication.

Issue #23, opened 2021-01-28
glass: create a login page


s0nea:
Create an entry page for the user in order to log in.

Required elements:
- Input field for the username (text)
- Input field for the password (password)
- Button 'Login'

Optional elements:
- Project logo
- Link to project website and/or docs (if available)

Related to:
* https://github.com/aquarist-labs/aquarium/issues/44
* https://github.com/aquarist-labs/aquarium/issues/26

PR #578, opened 2021-06-18
gravel/glass: user management

votdev:
The default credentials set up during deployment are: admin/aquarium

Peek 2021-06-25 17-27

Peek 2021-06-25 17-40

Fixes: https://github.com/aquarist-labs/aquarium/issues/421
Fixes: https://github.com/aquarist-labs/aquarium/issues/23
Fixes: https://github.com/aquarist-labs/aquarium/issues/26

## Checklist

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>

### DIFF
--- a/src/aquarium.py
+++ b/src/aquarium.py
@@ -39,12 +39,14 @@ from gravel.api import (
     local,
     devices,
     nfs,
+    auth,
+    user,
 )
 
 logger: logging.Logger = fastapi_logger
 
 
-async def aquarium_startup(aquarium_app: FastAPI, aquarium_api: FastAPI):
+async def aquarium_startup(_: FastAPI, aquarium_api: FastAPI):
     lvl = "INFO" if not os.getenv("AQUARIUM_DEBUG") else "DEBUG"
     setup_logging(lvl)
     logger.info("Aquarium startup!")
@@ -100,7 +102,7 @@ async def aquarium_startup(aquarium_app: FastAPI, aquarium_api: FastAPI):
     aquarium_api.state.nodemgr = nodemgr
 
 
-async def aquarium_shutdown(aquarium_app: FastAPI, aquarium_api: FastAPI):
+async def aquarium_shutdown(_: FastAPI, aquarium_api: FastAPI):
     logger.info("Aquarium shutdown!")
     await aquarium_api.state.gstate.shutdown()
     logger.info("shutting down node manager")
@@ -136,6 +138,14 @@ def aquarium_factory(
         {
             "name": "devices",
             "description": "Obtain and perform operations on cluster devices"
+        },
+        {
+            "name": "auth",
+            "description": "Operations related to user authentication"
+        },
+        {
+            "name": "user",
+            "description": "Operations related to user management"
         }
     ]
 
@@ -163,6 +173,8 @@ def aquarium_factory(
     aquarium_api.include_router(nodes.router)
     aquarium_api.include_router(devices.router)
     aquarium_api.include_router(nfs.router)
+    aquarium_api.include_router(auth.router)
+    aquarium_api.include_router(user.router)
 
     #
     # mounts
--- a/src/glass/src/app/pages/login-page/login-page.component.html
+++ b/src/glass/src/app/pages/login-page/login-page.component.html
@@ -1,0 +1,18 @@
+<div class="glass-login-page">
+  <div class="glass-login-page-background"></div>
+  <div class="glass-login-page-content"
+       fxLayout="column">
+    <mat-card>
+      <mat-card-content>
+        <glass-declarative-form [config]="config"></glass-declarative-form>
+      </mat-card-content>
+      <mat-card-actions fxLayoutAlign="center center">
+        <glass-submit-button [form]="form?.formGroup"
+                             [formId]="config.id"
+                             (buttonClick)="onLogin()">
+          <span translate>Login</span>
+        </glass-submit-button>
+      </mat-card-actions>
+    </mat-card>
+  </div>
+</div>
--- a/src/glass/src/app/pages/users-page/users-page.component.ts
+++ b/src/glass/src/app/pages/users-page/users-page.component.ts
@@ -1,0 +1,210 @@
+import { Component } from '@angular/core';
+import { marker as TEXT } from '@biesbjerg/ngx-translate-extract-marker';
+import { BlockUI, NgBlockUI } from 'ng-block-ui';
+import { finalize } from 'rxjs/operators';
+
+import { DeclarativeFormModalComponent } from '~/app/core/modals/declarative-form/declarative-form-modal.component';
+import { translate } from '~/app/i18n.helper';
+import { DialogComponent } from '~/app/shared/components/dialog/dialog.component';
+import { DatatableActionItem } from '~/app/shared/models/datatable-action-item.type';
+import { DatatableColumn } from '~/app/shared/models/datatable-column.type';
+import { DatatableData } from '~/app/shared/models/datatable-data.type';
+import { DeclarativeFormModalConfig } from '~/app/shared/models/declarative-form-modal-config.type';
+import { User, UserService } from '~/app/shared/services/api/user.service';
+import { DialogService } from '~/app/shared/services/dialog.service';
+
+@Component({
+  selector: 'glass-users-page',
+  templateUrl: './users-page.component.html',
+  styleUrls: ['./users-page.component.scss']
+})
+export class UsersPageComponent {
+  @BlockUI()
+  blockUI!: NgBlockUI;
+
+  loading = false;
+  firstLoadComplete = false;
+  data: User[] = [];
+  columns: DatatableColumn[];
+
+  constructor(private dialogService: DialogService, private userService: UserService) {
+    this.columns = [
+      {
+        name: TEXT('Name'),
+        prop: 'username',
+        sortable: true
+      },
+      {
+        name: TEXT('Full Name'),
+        prop: 'full_name',
+        sortable: true
+      },
+      {
+        name: TEXT('Disabled'),
+        prop: 'disabled',
+        sortable: true,
+        cellTemplateName: 'checkIcon'
+      },
+      {
+        name: '',
+        prop: '',
+        cellTemplateName: 'actionMenu',
+        cellTemplateConfig: this.onActionMenu.bind(this)
+      }
+    ];
+  }
+
+  loadData(): void {
+    this.loading = true;
+    this.userService.list().subscribe((data) => {
+      this.data = data;
+      this.loading = this.firstLoadComplete = true;
+    });
+  }
+
+  onAdd(): void {
+    this.dialogService.open(
+      DeclarativeFormModalComponent,
+      (res: User | boolean) => {
+        if (res) {
+          this.blockUI.start(translate(TEXT('Please wait, creating user ...')));
+          this.userService
+            .create(res as User)
+            .pipe(finalize(() => this.blockUI.stop()))
+            .subscribe(() => {
+              this.loadData();
+            });
+        }
+      },
+      {
+        width: '40%',
+        data: {
+          title: TEXT('Add User'),
+          fields: [
+            {
+              type: 'text',
+              label: TEXT('Name'),
+              name: 'username',
+              value: '',
+              validators: {
+                required: true
+              }
+            },
+            {
+              type: 'text',
+              label: TEXT('Full Name'),
+              name: 'full_name',
+              value: ''
+            },
+            {
+              type: 'password',
+              label: TEXT('Password'),
+              name: 'password',
+              value: '',
+              validators: {
+                required: true
+              }
+            },
+            {
+              type: 'checkbox',
+              label: TEXT('Disabled'),
+              name: 'disabled',
+              value: false,
+              hint: TEXT('Temporarily prohibit the user from logging in.')
+            }
+          ],
+          submitButtonText: TEXT('Add')
+        } as DeclarativeFormModalConfig
+      }
+    );
+  }
+
+  onActionMenu(user: User): DatatableActionItem[] {
+    const result: DatatableActionItem[] = [
+      {
+        title: TEXT('Edit'),
+        callback: (data: DatatableData) => {
+          this.dialogService.open(
+            DeclarativeFormModalComponent,
+            (res: User | boolean) => {
+              if (res) {
+                this.blockUI.start(translate(TEXT('Please wait, updating user ...')));
+                this.userService
+                  .update((res as User).username, res as User)
+                  .pipe(finalize(() => this.blockUI.stop()))
+                  .subscribe(() => {
+                    this.loadData();
+                  });
+              }
+            },
+            {
+              width: '40%',
+              data: {
+                title: TEXT('Edit User'),
+                fields: [
+                  {
+                    type: 'text',
+                    label: TEXT('Name'),
+                    name: 'username',
+                    value: user.username,
+                    readonly: true
+                  },
+                  {
+                    type: 'text',
+                    label: TEXT('Full Name'),
+                    name: 'full_name',
+                    value: user.full_name
+                  },
+                  {
+                    type: 'password',
+                    label: TEXT('Password'),
+                    name: 'password',
+                    value: ''
+                  },
+                  {
+                    type: 'checkbox',
+                    label: TEXT('Disabled'),
+                    name: 'disabled',
+                    value: user.disabled,
+                    hint: TEXT('Temporarily prohibit the user from logging in.')
+                  }
+                ],
+                submitButtonText: TEXT('Edit')
+              } as DeclarativeFormModalConfig
+            }
+          );
+        }
+      },
+      {
+        title: TEXT('Delete'),
+        callback: (data: DatatableData) => {
+          this.dialogService.open(
+            DialogComponent,
+            (res: boolean) => {
+              if (res) {
+                this.blockUI.start(translate(TEXT('Please wait, deleting user ...')));
+                this.userService
+                  .delete(data.username)
+                  .pipe(finalize(() => this.blockUI.stop()))
+                  .subscribe(() => {
+                    this.loadData();
+                  });
+              }
+            },
+            {
+              width: '40%',
+              data: {
+                type: 'yesNo',
+                icon: 'question',
+                message: TEXT(
+                  `Do you really want to delete user <strong>${data.username}</strong>?`
+                )
+              }
+            }
+          );
+        }
+      }
+    ];
+    return result;
+  }
+}
--- a/src/glass/src/app/shared/components/submit-button/submit-button.component.ts
+++ b/src/glass/src/app/shared/components/submit-button/submit-button.component.ts
@@ -9,7 +9,10 @@ import * as _ from 'lodash';
 })
 export class SubmitButtonComponent implements OnInit {
   @Input()
-  form?: FormGroup;
+  formId?: string;
+
+  @Input()
+  form?: FormGroup | undefined;
 
   @Output()
   buttonClick = new EventEmitter<Event>();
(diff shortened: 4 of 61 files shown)


src/aquarium.py
```diff
@@ -39,12 +39,17 @@
...
+    auth,
+    user,
 )
 
 logger: logging.Logger = fastapi_logger
 
 
 async def aquarium_startup(aquarium_app: FastAPI, aquarium_api: FastAPI):
+    import pydevd_pycharm
+    pydevd_pycharm.settrace('192.168.178.14', port=22222, stdoutToServer=True, stderrToServer=True)
```
jecluis, line 51: This should be conditional, somehow. Maybe passing an environment variable or something.

src/gravel/api/auth.py
```diff
@@ -0,0 +1,47 @@
...
+    tags=["auth"]
+)
+
+
+class LoginReplayModel(BaseModel):
+    access_token: str = Field(title="The JWT access token")
+
+
+@router.post("/login", response_model=LoginReplayModel)
+async def login(credentials: CredentialsModel, request: Request, authorize: AuthJWT = Depends()):
```
jecluis, line 37: specify the return type

src/gravel/api/auth.py
```diff
@@ -0,0 +1,47 @@
...
+    authenticated = await request.app.state.user_mgr.authenticate(credentials)
+    if not authenticated:
+        raise HTTPException(status_code=401, detail="Bad username or password")
+    access_token = authorize.create_access_token(subject=credentials.username)
+    return LoginReplayModel(access_token=access_token)
+
+
+@router.post("/logout")
+def login():
```
jecluis, line 46: s/login/logout/

src/gravel/api/user.py
```diff
@@ -0,0 +1,48 @@
...
+logger: Logger = fastapi_logger
+
+router: APIRouter = APIRouter(
+    prefix="/user",
+    tags=["user"]
+)
+
+
+@router.get("/", name="Get list of users", response_model=List[UserModel])
+async def enumerate_users(request: Request):
```
jecluis, line 32: specify return type

src/gravel/api/user.py
```diff
@@ -0,0 +1,48 @@
...
+    return await request.app.state.gstate.user_mgr.enumerate()
+
+
+@router.put("/", name="Create a new user")
+async def create_user(user: UserModel, request: Request):
```
jecluis, line 37: specify return type

src/gravel/api/user.py
```diff
@@ -0,0 +1,48 @@
...
+    return await request.app.state.gstate.user_mgr.add(user)
+
+
+@router.get("/{username}", name="Get a user by name")
+async def get_user(username: str, request: Request):
```
jecluis, line 42: return type

src/gravel/api/user.py
```diff
@@ -0,0 +1,48 @@
...
+    return await request.app.state.gstate.user_mgr.get(username)
+
+
+@router.delete("/{username}", name="Delete a user by name")
+async def delete_user(username: str, request: Request):
```
jecluis, line 47: return type

src/gravel/controllers/user.py
```diff
@@ -0,0 +1,74 @@
...
+class CredentialsModel(BaseModel):
+	username: str = Field("", title="The user name")
+	password: str = Field("", title="The password in plain text")
+
+
+class UserModel(BaseModel):
+	username: str = Field("", title="The user name")
+	password: str = Field("", title="The password")
+	full_name: str = Field("", title="The full name of the user")
+	disabled: Optional[bool] = Field(False, title="Is the user disabled?")
```
jecluis, line 32: if we're setting the default as `False`, I don't see a reason to keep this as `Optional`. If the field is not specified, it will take the default value.

src/gravel/controllers/user.py
```diff
@@ -0,0 +1,74 @@
...
+		return result
+
+
+class UserMgr:
+	def __init__(self, kv: KV):
+		self._kv = kv
+
+	async def enumerate(self) -> List[UserModel]:
+		values = await self._kv.get_prefix('/user/')
+		return [UserModel.parse_obj(json.loads(value)) for value in values]
```
jecluis, line 49: This could be a `parse_raw()` instead of having to do a `json.loads()` and a `parse_obj()`, I think.

src/gravel/controllers/user.py
```diff
@@ -0,0 +1,74 @@
...
+
+	async def exists(self, username: str) -> bool:
+		user = await self.get(username)
+		return user is not None
+
+	async def get(self, username: str) -> Optional[UserModel]:
+		value: str = await self._kv.get(f'/user/{username}')
+		if value is None:
+			return None
+		user = UserModel.parse_obj(json.loads(value))
```
jecluis, line 59: same here, I think this could be a `parse_raw()`.

src/gravel/api/user.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@router.put("/", name="Create a new user")
+async def create_user(user: UserModel, request: Request):
+    return await request.app.state.gstate.user_mgr.add(user)
+
+
+@router.get("/{username}", name="Get a user by name")
+async def get_user(username: str, request: Request):
+    return await request.app.state.gstate.user_mgr.get(username)
```
jecluis, line 43: This might be returning "None", which is something we probably want to check and convey something more descriptive to the user, even if by raising an exception with a status code.

jecluis (changes requested):
few things of note, and we need tests for this.

src/aquarium.py
jecluis: Also, this is a dependency that is not being added. Which I support; I don't think adding this dependency is something we want. So maybe do this if the dependency exists, like in a `try` block of the sorts of

```
try:
  import pydevd_pycharm
  pydevd_pycharm.settrace(...)
except ModuleNotFound:
  pass
```

votdev: This will be removed in the final PR.

src/gravel/controllers/user.py
votdev: > This could be a parse_raw() instead of having to do a json.loads() and a parse_obj(), I think.

Right, according to https://pydantic-docs.helpmanual.io/usage/models/#helper-functions this is correct.

votdev:
jenkins run tumbleweed

votdev:
jenkins run leap

votdev:
jenkins test tumbleweed

votdev:
jenkins test leap

votdev:
jenkins test tumbleweed

votdev:
> Not sure if it's part of this PR, but the red notification which pops up after a user has logged out is not something I'd prefer to show a user on successful logout. Due to the red color it looks like an error and the message seems to be more helpful to developers than to users. Maybe it's better to omit it?

Fixed

src/glass/src/app/shared/services/auth-storage.service.ts
```diff
@@ -0,0 +1,39 @@
...
+import * as _ from 'lodash';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class AuthStorageService {
+  constructor() {}
+
+  set(username: string): void {
+    localStorage.setItem('username', username);
```
jecluis, line 25: where is `localStorage` coming from?

src/gravel/api/auth.py
```diff
@@ -0,0 +1,56 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+from logging import Logger
+from fastapi.logger import logger as fastapi_logger
+from fastapi.routing import APIRouter
+from fastapi import HTTPException, Request, Response, Depends, status
+from fastapi.security import OAuth2PasswordRequestForm
+from pydantic import BaseModel, Field
+from gravel.controllers.auth import jwt_auth_scheme, JWTDenyList, JWTManager, JWT, UserMgr
```
jecluis, line 20: Please limit this at 80 cols, and split  them into lines in the format
```
from gravel.controllers.auth import (
  jwt_auth_scheme,
  ...
)
```

src/gravel/api/auth.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class LoginReplyModel(BaseModel):
+    access_token: str = Field(title="The access token")
+    token_type: str = Field("bearer")
+
+
+@router.post("/login", response_model=LoginReplyModel)
+async def login(request: Request, response: Response,
+                form_data: OAuth2PasswordRequestForm = Depends()) -> LoginReplyModel:
```
jecluis, line 38: I'd prefer the return to be moved to the next line and properly limit the length at 80 cols. Something like

```
async def login(request: Request, response: Response,
                form_data: OAuth2PasswordRequestForm = Depends()
) -> LoginReplyModel:
```

src/gravel/controllers/auth.py
```diff
@@ -0,0 +1,207 @@
...
+        user: Optional[UserModel] = await self.get(username)
+        if user is None:
+            return False
+        if user.disabled:
+            return False
+        return user.verify_password(password)
+
+
+class JWT(NamedTuple):
+    iss: str
```
jecluis, line 81: mind adding comments explaining what each one of these things are?

src/gravel/controllers/auth.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def includes(self, token: JWT) -> bool:
+        return token.jti in self._jti_dict
+
+
+class JWTAuthSchema(OAuth2PasswordBearer):
+    def __init__(self):
+        super().__init__(tokenUrl="auth/login")
+
+    async def __call__(self, request: Request) -> Optional[JWT]:  # type: ignore[override]
```
jecluis, line 169: The API is leaking here by passing a `Request` to the controller. It would be preferable to pass what this class actually requires (seems it's the `GlobalState`?) instead of a fastapi request.

src/gravel/controllers/auth.py
```diff
@@ -0,0 +1,207 @@
...
+            # Fallback: Try to get it from the headers.
+            token = await super().__call__(request)
+        # Decode token and do the following checks:
+        jwt_mgr = JWTManager(state.gstate.config.options.auth)
+        raw_token: JWT = jwt_mgr.get_raw_access_token(token)
+        # - Is the token revoked?
+        deny_list = JWTDenyList(state.gstate.store)
+        await deny_list.load()
+        if deny_list.includes(raw_token):
+            raise HTTPException(
```
jecluis, line 186: Raise a controller specific exception, and translate that to whatever is the appropriate HTTPException in the API controller.

src/gravel/controllers/auth.py
```diff
@@ -0,0 +1,207 @@
...
+                status_code=status.HTTP_401_UNAUTHORIZED,
+                detail="Token has been revoked"
+            )
+        # - Does the user exist?
+        user_mgr = UserMgr(state.gstate.store)
+        user: Optional[UserModel] = await user_mgr.get(str(raw_token.sub))
+        if user is None:
+            raise HTTPException(
```
jecluis, line 194: same here, and below.

jecluis (changes requested):
Few comments. In terms of style, mostly have issues with line length and line breaks (or lack of), but given we are not enforcing them I'm not going to hold the PR on that account.

We are leaking a lot of API state and structures to the core controllers, which should not happen, as it will make unit testing harder, and from an architectural point of view it makes little sense IMO.

Also, we're lacking unit tests for the added structures. We'll need tests.

src/aquarium.py
```diff
@@ -13,7 +13,6 @@
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU General Public License for more details.
 
-import logging
```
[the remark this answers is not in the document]
votdev: Right, strange, my IDE suggests to remove it.

src/glass/src/app/shared/services/auth-storage.service.ts
votdev: It's global available. See https://developer.mozilla.org/de/docs/Web/API/Window/localStorage

votdev: You can also use `window.localStorage`.

src/aquarium.py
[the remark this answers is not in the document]
votdev: Do we really need `import logging.config`?

src/gravel/api/auth.py
votdev: > I'd prefer the return to be moved to the next line and properly limit the length at 80 cols. Something like
> 
> ```
> async def login(request: Request, response: Response,
>                 form_data: OAuth2PasswordRequestForm = Depends()
> ) -> LoginReplyModel:
> ```

My IDE does not like that and shows a PEP8 error. Instead we should format it this way:
```
async def login(
        request: Request, response: Response,
        form_data: OAuth2PasswordRequestForm = Depends()
) -> LoginReplyModel:
```

src/aquarium.py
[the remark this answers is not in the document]
jecluis: `logging.config`, I don't think so.

src/gravel/api/auth.py
jecluis: I most definitely agree. I just didn't want to go and be too annoying :)

src/gravel/controllers/auth.py
votdev: The method signature is given by the derived class [OAuth2PasswordBearer](https://github.com/tiangolo/fastapi/blob/master/fastapi/security/oauth2.py#L152)

Can we relocate this class to somewhere else in our source tree where we do not pollute our `controller` rules?

src/gravel/controllers/auth.py
votdev: That's not possible, an instance of this class is injected in every API endpoint and is processed by FastAPI internally. No way to catch and map it.

src/gravel/controllers/auth.py
jecluis: If this is bound to the API, then it should probably live under `api/auth.py`.

src/gravel/controllers/auth.py
jecluis: By moving this to the `api/auth.py` file, we get rid of this annoyance and we raising HTTPExceptions still make sense IMO.

s0nea:
I just setup a testing environment for the PR. While doing this two questions came into my mind:

-  Would it be sensible to print the default username/password by e.g. the `run_aquarium.sh` script?
or
- If it's possible: is there a plan to define the default credentials while running the install wizard?

votdev:
>     * If it's possible: is there a plan to define the default credentials while running the install wizard?

As far as i remember we came to the conclusion to do not request the username and password during the installation.

jecluis:
> I just setup a testing environment for the PR. While doing this two questions came into my mind:
> 
>     * Would it be sensible to print the default username/password by e.g. the `run_aquarium.sh` script?

I think this makes little sense given the user won't be able to see the console.

>     * If it's possible: is there a plan to define the default credentials while running the install wizard?

And I'd think that's a "yes", and a "must have". The user should be able to configure a username and a password during initial deployment, and even be able to provide an ssh pubkey to be able to ssh into the nodes (maybe?).

votdev:
> > I just setup a testing environment for the PR. While doing this two questions came into my mind:
> > ```
> > * Would it be sensible to print the default username/password by e.g. the `run_aquarium.sh` script?
> > ```
> 
> I think this makes little sense given the user won't be able to see the console.
> 
> > ```
> > * If it's possible: is there a plan to define the default credentials while running the install wizard?
> > ```
> 
> And I'd think that's a "yes", and a "must have". The user should be able to configure a username and a password during initial deployment, and even be able to provide an ssh pubkey to be able to ssh into the nodes (maybe?).

:-) Two different opinions. 

OK, i can add a username and password page to the `Create` wizard. Th eSSH pubkey should be done in a separate issue/PR.

s0nea:
> > Would it be sensible to print the default username/password by e.g. the run_aquarium.sh script?
> 
> The displayed information will be obsolete as soon as the password has been changed or the user jas been replaced by another one. Because of that it doesn't make sense to display it.

I thought about it as well. But given the fact the script is mostly used for development and testing it's worth it in my opinion.

s0nea:
> > I just setup a testing environment for the PR. While doing this two questions came into my mind:
> > ```
> > * Would it be sensible to print the default username/password by e.g. the `run_aquarium.sh` script?
> > ```
> 
> I think this makes little sense given the user won't be able to see the console.

It was just an idea because I setup the environment and then I had to look up the credentials in the code.

jecluis:
> OK, i can add a username and password page to the `Create` wizard. The SSH pubkey should be done in a separate issue/PR.

I think both can/should be a different PR. We might not want to hold this one right now. Also, this one still needs tests to be merged, even if basic ones.

src/glass/src/app/pages/login-page/login-page.component.html
```diff
@@ -0,0 +1,17 @@
...
+  <div class="glass-login-page-background"></div>
+  <div class="glass-login-page-content"
+       fxLayout="column">
+    <mat-card>
+      <mat-card-content>
+        <glass-declarative-form [config]="config"></glass-declarative-form>
+      </mat-card-content>
+      <mat-card-actions fxLayoutAlign="center center">
+        <glass-submit-button [form]="form?.formGroup"
+                             (buttonClick)="onLogin()">
```
s0nea, line 11: Would it be possible to call `doLogin()` by hitting 'enter' after entering the password? Just a minor thing. It shouldn't block the PR from being merged and can be addressed in a different issue if you prefer.

src/glass/src/app/pages/users-page/users-page.component.ts
```diff
@@ -0,0 +1,210 @@
...
+                  }
+                ],
+                submitButtonText: TEXT('Edit')
+              } as DeclarativeFormModalConfig
+            }
+          );
+        }
+      },
+      {
+        title: TEXT('Delete'),
```
s0nea, line 179: Can also be addressed in a separate issue: I do have the feeling we should prevent the logged-in user from deleting himself.

src/glass/src/app/pages/login-page/login-page.component.html
votdev: No, that's not possible because the submit button is not part of the form. If this is required, we need to refactor the form implementation.

votdev: I'll check that, didn't find the reason why it does not work till now.

src/glass/src/app/pages/users-page/users-page.component.ts
votdev: The API endpoint will check that.

src/glass/src/app/pages/login-page/login-page.component.html
votdev: Fixed

s0nea: Nice :partying_face: Thank you!

votdev:
jenkins test tumbleweed

src/glass/src/app/pages/users-page/users-page.component.ts
s0nea: :+1: I think it would be great if the user **can't** disable himself in addition.

s0nea:
Somehow I can't re-enable a deactivated user:

https://user-images.githubusercontent.com/8761082/124268524-e7973f80-db39-11eb-9e89-a2c2779e98a5.mp4

Any ideas?

votdev:
> Somehow I can't re-enable a deactivated user:
> aqua.mp4
> 
> Any ideas?

Fixed

s0nea:
jenkins test leap

src/glass/src/app/shared/components/submit-button/submit-button.component.ts
```diff
@@ -20,19 +23,16 @@ export class SubmitButtonComponent implements OnInit {
...
-        // Process all invalid controls and update them to draw the
-        // as invalid.
-        _.forEach<Record<string, AbstractControl>>(
-          this.form.controls,
-          (control: AbstractControl, key: string) => {
-            if (control.invalid) {
-              control.markAllAsTouched();
-              control.updateValueAndValidity();
-            }
+      // Process all invalid controls and update them to draw the
```
p-se, line 26: incomplete comment?

votdev: Fixed

votdev:
jenkins test tumbleweed

s0nea (approved):
The `glass` part LGTM!

I suggest to wait for @jecluis's final review before merging the PR.

jecluis (approved):

PR #580, opened 2021-06-20
gravel: support system disk on joining node, assimilate devices individually (part 2)

jecluis:
With this patch set we now support system disk creation on a joining node, as well as assimilating devices individually (instead of a blanket `all available`), both on a seed node and on a joining node.

~This patch set depends on merging #575  and  #581~

### DIFF
--- a/src/gravel/controllers/nodes/deployment.py
+++ b/src/gravel/controllers/nodes/deployment.py
@@ -307,7 +307,6 @@ class NodeDeployment:
         conn = await self._connmgr.connect(uri)
         logger.debug(f"join > conn: {conn}")
 
-        self._set_hostname(hostname)
         joinmsg = JoinMessageModel(
             uuid=uuid,
             hostname=hostname,
@@ -316,8 +315,6 @@ class NodeDeployment:
         )
         msg = MessageModel(type=MessageTypeEnum.JOIN, data=joinmsg.dict())
         await conn.send(msg)
-
-        self._state.mark_join()
 
         reply: MessageModel = await conn.receive()
         logger.debug(f"join > recv: {reply}")
@@ -338,13 +335,17 @@ class NodeDeployment:
         assert welcome.keyring
         assert welcome.etcd_peer
 
-        # create disks here
+        # create system disk after we are certain we are joining.
+        # ensure all state writes happen only after the disk has been created.
         systemdisk = SystemDisk(self._gstate)
         try:
             await systemdisk.create(disks.system)
             await systemdisk.enable()
         except GravelError as e:
             raise NodeCantJoinError(e.message)
+
+        self._state.mark_join()
+        self._set_hostname(hostname)
 
         my_url: str = \
             f"{hostname}=http://{address}:2380"
--- a/src/gravel/controllers/nodes/deployment.py
+++ b/src/gravel/controllers/nodes/deployment.py
@@ -38,6 +38,7 @@ from gravel.controllers.nodes.errors import (
     NodeAlreadyJoiningError,
     NodeBootstrappingError,
     NodeCantDeployError,
+    NodeCantJoinError,
     NodeError,
     NodeHasBeenDeployedError,
     NodeHasJoinedError,
@@ -280,7 +281,8 @@ class NodeDeployment:
         token: str,
         uuid: UUID,
         hostname: str,
-        address: str
+        address: str,
+        disks: DeploymentDisksConfig
     ) -> bool:
         logger.debug(
             f"join > with leader {leader_address}, token: {token}"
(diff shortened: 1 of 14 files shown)


jecluis:
jenkins test leap

src/gravel/controllers/nodes/systemdisk.py
```diff
@@ -84,7 +84,8 @@ class SystemDisk:
         "etc": "/etc",
         "logs": "/var/log",
         "etcd": "/var/lib/etcd",
-        "containers": "/var/lib/containers"
+        "containers": "/var/lib/containers",
+        "roothome": "/root"
```
mgfritch, line 88: `/root` is needed for orch ssh config? or something else?

src/gravel/controllers/orch/orchestrator.py
```diff
@@ -46,25 +46,47 @@ def host_ls(self) -> List[OrchHostListModel]:
...
 
-    def assimilate_all_devices(self) -> None:
+    def assimilate_devices(self, host: str, devices: List[str]) -> None:
+
+        devs: str = ','.join(devices)
         cmd = {
-            "prefix": "orch apply osd",
-            "all_available_devices": True
+            "prefix": "orch daemon add osd",
+            "svc_arg": f"{host}:{devs}"
```
mgfritch, line 65: this is a synchronous call in the orchestrator ... and will create an 'unmanaged' osd spec

should we pass the devs via a DriveGroupSpec to the async `orch apply osd` command?

src/gravel/controllers/nodes/systemdisk.py
jecluis: yeah, ssh pubkey

src/gravel/controllers/orch/orchestrator.py
jecluis: that explains why it takes so long to finish. I guess the alternative needs to be the spec :\

AvengerMoJo (commented):
Looks good to me.

jecluis:
@mgfritch fixed, using a drive group spec, and it seems to be working. Added pyyaml as a dependency though. Could have specified the spec as free-form text, but would rather get pyyaml in anyway because, I suspect, is going to be useful as we move forward (also, ensures it's actually yaml in the end :shrug: )

jecluis:
jenkins run tumbleweed

src/gravel/controllers/orch/orchestrator.py
```diff
@@ -36,35 +38,71 @@ class Orchestrator:
...
+            "data_devices": {
+                "paths": devices
+            }
+        }
+        specstr: Optional[str] = None
+        try:
+            specstr = yaml.dump(spec)  # type: ignore
+        except Exception as e:
+            raise GravelError(str(e))
+        assert specstr is not None
```
mgfritch, line 78:
```suggestion
        specstr: str = yaml.dump(spec)
```

mgfritch: I don't think this needs exception handling? It also appears as if `yaml.dump` will always produce a str type..

src/gravel/controllers/orch/ceph.py
```diff
@@ -150,13 +150,14 @@ def fsid(self) -> str:
...
-             cmd: Dict[str, Any]
+             cmd: Dict[str, Any],
+             inbuf: bytes = b""
              ) -> Any:
         self.assert_is_ready()
         cmdstr: str = json.dumps(cmd)
 
         try:
-            rc, out, outstr = func(cmdstr, b"")
+            rc, out, outstr = func(cmdstr, inbuf)
```
mgfritch, line 160:
```suggestion
            rc, out, outstr = func(cmdstr, inbuf.encode("utf-8'))
```

might be cleaner todo the bytes conversion here ?

mgfritch (dismissed):
mostly nits, but lgtm!

src/gravel/controllers/orch/orchestrator.py
jecluis: good call. It may return `None` however, so I'll keep the assert, even if it's just for static type checker purposes.

src/gravel/controllers/orch/ceph.py
jecluis: Personally, I think it's better to make it clear that these functions are taking `bytes` instead of string, and do the conversion as earlier in the call stack as possible. That matches the expectation of an actual `call()` to the underlying thing.

src/gravel/controllers/nodes/deployment.py
```diff
@@ -440,32 +477,29 @@ async def finish_bootstrap_cb(
...
                 )
-                logger.error("unable to assimilate devices")
-                logger.exception(e)
-                await finisher(False, str(e))
+                await finisher(False, e.message)
+            else:
+                self._progress = ProgressEnum.DONE
+                await finisher(True, None)
+
+        async def _assimilate_devices() -> None:
```
jhesketh, line 495: (nit) Might want to move this to before `finish_bootstrap_cb` since it is called from within there and scoping gets confusing with the identically named methods on the class.

Alternatively (and personal preference) this doesn't need to be a method and could just be inside of the above `try` statement for clarity.

jhesketh (approved):

PR #582, opened 2021-06-21
image: cleanup, drop ignition and combustion

jecluis:

### DIFF
--- a/images/aquarium/config.xml
+++ b/images/aquarium/config.xml
@@ -59,7 +59,7 @@
             bootloader="grub2"
             bootloader_console="gfxterm"
             firmware="uefi"
-            kernelcmdline="quiet systemd.show_status=yes console=ttyS0,115200 console=tty0 net.ifnames=0 \$ignition_firstboot ignition.platform.id=metal root=/dev/vda3 disk=/dev/vda"
+            kernelcmdline="quiet systemd.show_status=yes console=ttyS0,115200 console=tty0 net.ifnames=0 root=/dev/vda3 disk=/dev/vda"
             bootpartition="false"
             bootkernel="custom"
             devicepersistency="by-uuid"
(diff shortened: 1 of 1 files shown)


images/aquarium/config.xml
```diff
@@ -15,17 +15,19 @@
         <!-- Flavors -->
 	    <!-- <profile name="ContainerHost" description="Aquarium with container runtime (podman)"/> -->
         <!-- Platforms -->
-        <profile name="Vagrant-VirtualBox-x86_64" description="Vagrant for x86_64 VirtualBox" arch="x86_64"/>
-        <profile name="Vagrant-x86_64" description="Vagrant for x86_64 libvirt" arch="x86_64"/>
+        <profile name="Vagrant-VirtualBox-x86_64" description="Vagrant for
+        x86_64 with VirtualBox" arch="x86_64"/>
+        <profile name="Vagrant-x86_64" description="Vagrant for x86_64 with libvirt" arch="x86_64"/>
```
kshtsk, line 20: I wonder if this change is really needed and if so, can put new lines between attributes instead of inside attribute values. Also it would be nice to have some indentation so the start of the tag is distinguished.

images/aquarium/config.xml
```diff
@@ -15,17 +15,19 @@
...
         <profile name="Vagrant" description="Vagrant" arch="x86_64">
             <requires profile="Vagrant-x86_64"/>
         </profile>
         <profile name="SelfInstall" description="Self Installing Image" arch="x86_64"/>
         <!-- Images (flavor + platform) -->
-        <profile name="Vagrant-VirtualBox" description="MicroOS with Podman for Vagrant" arch="x86_64">
+        <profile name="Vagrant-VirtualBox" description="Vagrant
+        with VirtualBox" arch="x86_64">
```
kshtsk, line 27: See comment above.

images/aquarium/config.xml
jecluis: It is not _needed_, but the wording was annoying me enough that I did it anyway. Good point with the wrapping and indenting.

images/aquarium/config.xml
jecluis: You are right, this is ugly. Didn't notice that with the editor for some reason :shrug:

kshtsk:
jenkins run ubuntu

kshtsk (approved):
looks good now

AvengerMoJo (approved):
Look good to me. Thanks

PR #585, opened 2021-06-22
images: turn 'self-install' into a ramdisk

jecluis:


AvengerMoJo (approved):
Looks good to me

PR #603, opened 2021-07-05
gravel: ctrl/network: keep track of interfaces (part 1 of many)

jecluis:
This patch set allows us to keep track of network interfaces and their statistics.

It will be a part of a longer series that will lead us to being able to configure network interfaces from the installer.


jecluis:
jenkins test leap

src/aquarium.py
```diff
@@ -92,6 +93,11 @@ async def aquarium_startup(aquarium_app: FastAPI, aquarium_api: FastAPI):
         gstate.config.options.services.probe_interval, gstate, nodemgr)
     gstate.add_services(services)
 
+    network: NetworkInterfaces = NetworkInterfaces(
```
jhesketh, line 96: (nit) should this be `networks`?

src/gravel/controllers/gstate.py
```diff
@@ -169,6 +173,10 @@ def add_services(self, services: Services):
         self.services = services
         self.add_ticker('services', services)
 
+    def add_network(self, network: NetworkInterfaces):
```
jhesketh, line 176: Similarly (`add_network_interfaces`)?

jhesketh (approved):
lgtm, just a really minor nit that I think can be ignored

src/aquarium.py
votdev: I vote for `network_interfaces` or `network_ifaces`.

src/gravel/controllers/resources/network.py
```diff
@@ -0,0 +1,328 @@
...
+
+    def _get_devices(self) -> List[str]:
+        """ Obtain all PCI devices from /sys/bus/pci """
+        devpath = Path(self.devpath)
+        assert devpath.exists() and devpath.is_dir()
+        return next(os.walk(self.devpath))[1]
+
+    async def _get_pci_info(self, pcidev: str) -> PCIModel:
+        """ Obtain specific PCI device information, using lspci. """
+        cmd = f"lspci -vmm -s {pcidev}"
```
votdev, line 122: Why not using [pyudev](https://pyudev.readthedocs.io/en/latest/) instead of running CLI commands and parsing their output?

jecluis: because I wasn't aware of it -- thanks. Regardless, I think the path forward is relying on `lshw` because it provides a lot more information in one go about the network interfaces.

votdev:
Will a follow up PR include API endpoint to get these information via REST?

jecluis:
Yeah, I wanted to do a multi-part patchset to introduce things incrementally. But things got stuck for reasons, and this actually needs to be refactored a bit...

kshtsk:
jenkins run tumbleweed

kshtsk:
jenkins run tumbleweed

kshtsk:
jenkins run tumbleweed

kshtsk:
jenkins run tumbleweed

src/gravel/controllers/resources/ifaces.py
```diff
@@ -138,6 +138,26 @@ class InterfaceStatsModel(BaseModel):
     ts: int = Field(0, title="Reading timestamp")
 
 
+def _get_speed(v: Optional[str]) -> int:
+    """Obtain speed in bytes from string"""
```
jhesketh, line 142: Isn't this obtaining it in bits, not bytes?

Issue #607, opened 2021-07-09
gravel/glass: user management - prevent the current user from disabling himself


s0nea:
The logged in user can disable himself via the user management, currently. We should prevent him from doing this (similar as we prevent him from deleting himself).

PR #610, opened 2021-07-12
gravel/glass: user management - prevent the current user from disabling himself

votdev:
Handle this behaviour completely in the backend, no need to do any checks in the frontend.

Fixes: https://github.com/aquarist-labs/aquarium/issues/607

## Checklist
- [x] Includes tests for new functionality or reproducer for bug

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>

### DIFF
--- a/src/gravel/api/user.py
+++ b/src/gravel/api/user.py
@@ -77,12 +77,15 @@ async def delete_user(username: str,
 async def patch_user(username: str,
                      update_user: UserModel,
                      request: Request,
-                     _=Depends(jwt_auth_scheme)) -> UserModel:
+                     token: JWT = Depends(jwt_auth_scheme)) -> UserModel:
     user_mgr = UserMgr(request.app.state.gstate.store)
     user = await user_mgr.get(username)
     if user is None:
         raise HTTPException(status.HTTP_404_NOT_FOUND,
                             detail="User does not exist")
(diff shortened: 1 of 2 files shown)


s0nea (approved):

Devp00l (approved):
I would assume due to your test that the coverage should increase instead of decrease. Maybe we have a problem with the test runner? Nevertheless nice work :)

votdev:
> I would assume due to your test that the coverage should increase instead of decrease. Maybe we have a problem with the test runner? Nevertheless nice work :)

I think this is normal because some new code has added to gravel, but everything in api/ is not counted by the coverage + i assume that the API endpoint tests are not counted by the coverage, too.

jecluis:
> > I would assume due to your test that the coverage should increase instead of decrease. Maybe we have a problem with the test runner? Nevertheless nice work :)
> 
> I think this is normal because some new code has added to Gravel, but everything in `api/` is not counted by the coverage + i assume that the API endpoint tests are not counted by the coverage, too.

The code in `api/` is counted towards coverage, we just haven't been adding tests for it.

jecluis (approved):
lgtm

jhesketh (approved):

PR #616, opened 2021-07-19
[tools] Aqrdev add virtualbox provider and update the all the libaqr with the provider options

AvengerMoJo:
Adding VirtualBox support to the build. 
Adding Provider information by replacing the image type.
It should not change the usage compare to before. Listing the box and image now includes the provider type. 

Signed-off-by:  Alex Lau (AvengerMoJo) <alau@suse.com>


AvengerMoJo:
I think I did something wrong in the rebase, let me do that again.

PR #621, opened 2021-07-26
tools: Adds force option on box destroy

Devp00l:
This is helpful if aqrdev is used inside a script where user approval
just delays.

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>

### DIFF
--- a/tools/libaqua/cli.py
+++ b/tools/libaqua/cli.py
@@ -392,9 +392,10 @@ def cmd_start(ctx: AppCtx, name: str, conservative: bool) -> None:
 
 
 @app.command("stop")
+@click.option("-f", "--force", flag_value=True)
 @click.argument("name", required=True, type=str)
 @pass_appctx
-def cmd_stop(ctx: AppCtx, name: str) -> None:
+def cmd_stop(ctx: AppCtx, name: str, force: bool) -> None:
     """
     Stop deployment; destroys running machines
     """
@@ -402,7 +403,7 @@ def cmd_stop(ctx: AppCtx, name: str) -> None:
     assert deployment
 
     try:
-        deployment.stop()
+        deployment.stop(force=force)
     except AqrError as e:
         click.secho(f"Error: {e.message}", fg="red")
         sys.exit(e.errno)
--- a/tools/libaqua/deployment.py
+++ b/tools/libaqua/deployment.py
@@ -153,13 +153,13 @@ class Deployment:
         except AqrError as e:
             raise e
 
-    def stop(self, interactive: bool = True) -> None:
+    def stop(self, force: bool = False) -> None:
         """ Stop deployment """
         try:
             with vagrant.deployment(self._path) as deployment:
                 if deployment.shutoff or deployment.notcreated:
                     return  # nothing to do.
-                deployment.stop(interactive=interactive)
+                deployment.stop(force=force)
         except AqrError as e:
             raise e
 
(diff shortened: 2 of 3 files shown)


tools/aqrdev
```diff
@@ -348,17 +348,18 @@ def cmd_start(ctx: AppCtx, name: str, conservative: bool) -> None:
...
+def cmd_stop(ctx: AppCtx, name: str, force: bool) -> None:
     """
     Stop deployment; destroys running machines
     """
     deployment: Deployment = _get_deployment(ctx, name)
     assert deployment
 
     try:
-        deployment.stop()
+        deployment.stop(not force)
```
votdev, line 362: What do you think about using this?
```suggestion
        deployment.stop(interactive=not force)
```

kshtsk: for better readability I suggest to use:
```
if force:
  deployment.stop(interactive=false)
else:
  deployment.stop()

**Comment by kshtsk** [comment]
jenkins run opensuse

File: tools/aqrdev [hunk 2]
```diff
@@ -348,17 +348,18 @@ def cmd_start(ctx: AppCtx, name: str, conservative: bool) -> None:
 
 
 @app.command("stop")
+@click.option("-f", "--force", flag_value=True)
```
kshtsk, line 351: Just wondering why 'force'? Maybe '--interactive' with defaults True, is better?

kshtsk: Alternatively, '-n', '--non-interactive'.

tools/aqrdev
kshtsk: Though, there is difference, non-interactive mode does abort if some question happens during execution, when force, it forces the answer with default option, so interactive is not the same as 'not force'.

tools/aqrdev
Devp00l: Sorry for the late response. Just thought ``--force`` is more common. I really didn't thought about interactive as in my head this sounds like something where someone can really be interactive with like a REPL and not just choose between yes or no which will result in execution or not.

Nevertheless I'm open to ideas and can change it to ``--interactive`` if this works better for most devs.

To keep it simple leave a thumbs up on my comment or a thumbs up on kyr's comment above. Thanks in advance :)

p-se (commented):
What about being consistent and naming both, the CLI switch and the internal variable either `force` or `interactive`? Would result in less confusion either way, IMHO.

```python
# Either
deployment.stop(interactive)
deployment.stop(interactive=interactive)
# or
deployment.stop(force)
deployment.stop(force=force)
```

tools/aqrdev
Devp00l: Changed the `interactive` variable in all used methods to the counterpart `force` in order to have consistency in the code base.

tools/aqrdev
Devp00l: @votdev used the suggested parameter assignment with the new `force` parameter.

Devp00l:
> What about being consistent and naming both, the CLI switch and the internal variable either `force` or `interactive`? Would result in less confusion either way, IMHO.
> 
> ```python
> # Either
> deployment.stop(interactive)
> deployment.stop(interactive=interactive)
> # or
> deployment.stop(force)
> deployment.stop(force=force)
> ```

Thanks Patrick :) Resolved the code confusion.

jecluis (approved):
lgtm, but let's merge this after #630

votdev (approved):

jecluis:
This now needs to be rebased against the changes of #630

jecluis (approved):

PR #622, opened 2021-07-26
tools: Use npm ci in setup-dev.sh

Devp00l:
### tools: Use npm ci in setup-dev.sh

Currently npm install is used which isn't correct as it doesn't
consider the versions mentioned in the `package-lock.json`,
`npm ci` however does.

This also updates the git workflow in order to test the right command.

---

### glass: Lock file version change
    
Fundamental changes in the lock file in order to use node 14 and older and not only 16.
    
This change also updates the packages.
    
---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>

### DIFF
--- a/tools/setup-dev.sh
+++ b/tools/setup-dev.sh
@@ -374,7 +374,6 @@ pip install -r src/requirements.txt -U || exit 1
 deactivate
 
 pushd src/glass &>/dev/null
-rm -rf node_modules
-npm install || exit 1
+npm ci || exit 1
 npx ng build || exit 1
 popd &>/dev/null
--- a/.github/workflows/glass-unit-tests.yml
+++ b/.github/workflows/glass-unit-tests.yml
@@ -9,14 +9,14 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [14.x, 15.x]
+        node-version: [14.x, 16.x]
     steps:
       - uses: actions/checkout@v2
       - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node-version }}
-      - run: npm install
+      - run: npm ci
         working-directory: src/glass
       - run: npx ng build
         working-directory: src/glass
--- a/src/glass/package-lock.json
+++ b/src/glass/package-lock.json
@@ -74,6 +74,19 @@
         "typescript": "~4.3.5"
       }
     },
+    "node_modules/@ampproject/remapping": {
+      "version": "1.0.1",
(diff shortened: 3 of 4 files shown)


s0nea:
jenkins test tumbleweed

Devp00l:
Hm.. seems like the npm version for leap and tumbleweed is to low. `npm ci` currently only works with node 16.x ... Not sure if this is an problem with our package lock file because it was updated at some point with node 16.x

kshtsk:
I guess we need to use 14.x version since not all tools support 16.x yet.

tools/setup-dev.sh
```diff
@@ -363,6 +363,6 @@ deactivate
 
 pushd src/glass &>/dev/null
 rm -rf node_modules
```
votdev, line 365: You can remove this line because `npm ci` deletes this directory also.

votdev (changes requested):

Devp00l:
Updated the PR description as I changed some things in order to make ``package-lock.json`` usable by older node versions, which is also tested now by the git hub actions.

tools/setup-dev.sh
Devp00l: Thanks for the hint :)

votdev (approved):
LGTM

kshtsk:
jenkins test tumbleweed

Devp00l:
@kshtsk which version is used in the jekins job as two nodejs versions are installed I assume the default node path will use node in version 5 (not sure but it comes with the packet ``nodejs-common-5.0-2.1.x86_64``) instead of 16 which is also installed. I tested it with node 14 up to 16 which worked fine.

kshtsk:
> @kshtsk which version is used in the jekins job as two nodejs versions are installed I assume the default node path will use node in version 5 (not sure but it comes with the packet `nodejs-common-5.0-2.1.x86_64`) instead of 16 which is also installed. I tested it with node 14 up to 16 which worked fine.

The version of nodejs can be found in the logs of the tests, also the way it is tested is well defined in .wasser/config.yaml what exactly installed on your system and how do you test it is unclear to me. So I don't know how did you test it with 16 if cypress does not support the 16?

kshtsk:
@Devp00l if you take a look at leap run logs, you can find that there is only nodejs installed which is 14.

kshtsk:
jenkins run tumbleweed

kshtsk:
The problem in regard which nodejs is taken on tumbleweed, because today it can be one version as default, another day it can be changed to another.

Devp00l:
Seems to have fixed itself ;)

PR #631, opened 2021-08-17
gravel: replace etcd with kvstore backed by ceph itself

tserong:
This is a bit rough.  There's all sorts of sharp corners you can cut your fingers on.  Also, I haven't actually removed the etcd bootstrap bits yet, and I haven't touched any of the tests.  Still, it does actually work. Well...  Almost.  I just tried adding a second node, and it's failing with "assert ntp_addr" in gravel/controllers/nodes/deployment.py's join() for some reason.  I'm not sure yet if that's due to this change or not. Still, feedback on the general approach would be greatly appreciated :-)

## Checklist

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>


tserong:
>  I just tried adding a second node, and it's failing with "assert ntp_addr" in gravel/controllers/nodes/deployment.py's join() for some reason. I'm not sure yet if that's due to this change or not

Having apparently thought about this further while I was asleep, I woke up this morning and realised this'll be because the second node can't talk to the cluster before it's joined (no ceph.conf & admin keyring), so of course won't be able to retrieve the ntp address from the kvstore. I guess I need to have the join return copies of these so the joining node can talk to the cluster immediately...

tserong:
Correction: ceph.conf and the admin keyring _are_ already there at that point (the join already does return those things). The problem is just that the cluster connection thread is inevitably in the middle of a 10 second sleep when those things land, so isn't connected when we try to get the ntp address.

tserong:
OK, 35eefa5 fixes the join case, but there's still something a bit janky about it that I haven't been able to nail down (see the comments in that commit)

tserong:
Please ignore the previous comment about jankiness.  Turns out what I was seeing was two calls to `ensure_connect()`, one from `GlobalState.init_store()` followed by one from `NodeDeployment.join()`. That's fine.

tserong:
This is now in a reasonable state for manual testing if anyone's keen. It works fine for cluster deployment and node join. I still expect it to potentially behave badly and/or lock up if the cluster goes dead later though.

tserong:
Rebased, also made black and mypy happy. Still need to fix the tests.

tserong:
jenkins run tumbleweed

tserong:
Rock 'n' roll.. I don't have unit tests for the new KV class yet, but at least I'm no longer breaking all the _other_ unit tests!

src/gravel/controllers/kv.py
```diff
@@ -12,125 +12,370 @@
...
+                        "kvstore",
+                        "# aquarium kv store is in this object's omap\n".encode(
+                            "utf-8"
+                        ),
+                    )
+                    # At this point, iff it's a new pool, new object, etc.
+                    # we need to push everything from our local cache to
+                    # the omap on our kvstore, to populate it with whatever
+                    # may have been set pre-bootstrap.
+                    # Probably want to lock the object here too, in case we get two
```
irq0, line 151: Would assert_version work instead of locking the object?
e.g `self._ioctx.assert_version(1)` before the set_omap below

src/gravel/controllers/kv.py
```diff
@@ -12,125 +12,370 @@
...
+                # we *know* this is a problem, or we just imagine it's a problem?)
 
-    async def get(self, key: str) -> Optional[str]:
-        """Get value for provided key"""
-        assert self._client
-        value, _ = await self._client.get(key)
+                with rados.WriteOpCtx() as op:
+                    self._ioctx.set_omap(op, (key,), (bvalue,))
+                    logger.debug("Doing write op")
+                    # write op naturally gets stuck if cluster badly degraded,
```
irq0, line 248: 1) Wondering if there is a way to hack the pg placement somehow to have more control :) Thinking out loud here, but would it make sense to scale the pool size to the number of aquarium nodes? 
Using the upmap feature[42] to pin the pg where we want it to be? As this map is quite important, perhaps even on every node?

2) Probably a naive question, but would it make sense to put the data into the cluster map or the Ceph config store? It would have the same availability as the mons, but updating is probably more expensive.

[42] https://docs.ceph.com/en/latest/rados/operations/balancer/#modes

src/gravel/controllers/kv.py
tserong: Maybe? I didn't know about that piece of functionality. Thanks for the tip :-)

src/gravel/controllers/kv.py
tserong: > 2\. Probably a naive question, but would it make sense to put the data into the cluster map or the Ceph config store? It would have the same availability as the mons, but updating is probably more expensive.

We previously discussed storing this stuff in the ceph config store. The only written reference I can find in a hurry is https://github.com/aquarist-labs/aquarium/issues/32, which is pretty terse. IIRC arguments against were:
* Practical: We can't access the data if the cluster is down, but of course the local cache fixes (or should fix) that, which kinda nullifies this point.
* Aesthetic: The ceph config store is for storing _ceph_ configuration, and aquarium is somehow "outside" ceph, or ceph is inside aquarium ;-) because aquarium is deploying it. From that perspective, we'd be "polluting" the ceph config store
* I can't remember if there were other concerns :-/ (@jecluis? @mgfritch?)

jecluis: > 1. Wondering if there is a way to hack the pg placement somehow to have more control :) Thinking out loud here, but would it make sense to scale the pool size to the number of aquarium nodes?
Using the upmap feature[42] to pin the pg where we want it to be? As this map is quite important, perhaps even on every node?

This would be overkill. It would be fine for a cluster up to 3 nodes or so, but then it becomes a liability because the pool would be degraded should a node die without anywhere to move the data to. Also, depending on what data we're storing, should we start storing lots of small writes (e.g., prometheus data), a single update would have to it all nodes in the cluster.

> I can't remember if there were other concerns :-/ (@jecluis? @mgfritch?)

Lack of watches, for one. Also, it would be nice if we could persist prometheus data in this pool. That way we don't need to have multiple instances of prometheus running, just the one and rely on Ceph for redundancy and persistence.

src/gravel/controllers/kv.py
```diff
@@ -12,125 +12,395 @@
...
+                    logger.debug(
+                        f"config watch id is {self._config_watch.get_id()}"
+                    )
+                    # will raise:
+                    # rados.ObjectNotFound: [errno 2] RADOS object not found (watch error)
+            except Exception as e:
+                # TODO: deal with this (log it? ignore it?)
+                # e.g. RADOS rados state (You cannot perform that operation on a Rados object in state configuring.)
+                # this makes the log pretty messy prior to bootstrap (errors every 10 seconds)
+                logger.error(str(e))
```
jecluis, line 175: Maybe handle exceptions on a per-operation basis and log those that make sense? Propagate when the operation should actually fail?

src/gravel/controllers/kv.py
```diff
@@ -12,125 +12,395 @@
...
+        self._run = False
+        self._event.set()
 
     async def put(self, key: str, value: str) -> None:
         """Put key/value pair"""
-        assert self._client
-        await self._client.put(key, value)
+        # Try to write to the kvstore in our pool, and also
+        # write to our local cache (whether or not the write
+        # to the pool succeeds)
```
jecluis, line 228: I think we should only write to the cache if the write succeeds, and ideally from the watches. If we're not watching for a specific key, I think it's okay to not cache it. This way we ensure that we always have a consistent state with the cluster, even if it gets out of sync when we die.

jecluis (commented):
Overall looks good, but I'm a bit worried about the caching semantics. Should it be feasible, I think we should not be updating the cache with values that we write, directly, but instead rely on watches to update cached values. I wonder whether that's appropriate as well should a value be deleted, because we need to clear up that value from our cache as well. Then again, that might be just a matter of always trying to read directly from omap, and only fallback to the cache if a cluster connection is not possible.

src/gravel/controllers/kv.py
```diff
@@ -12,125 +12,395 @@
...
+                logger.error(str(e))
+
+            # TODO: Should we sleep for longer?  A minute instead of 10 seconds?  This is pretty arbitrary...
+            logger.debug("Cluster connection thread sleeping for 10 seconds")
+            self._event.wait(10)
+            self._event.clear()
+            # How on earth to we detect that the cluster has gone away
+            # and trigger a reconnect?
+            # What happens if we try a write op on a down cluster?
+            # might want watch checks?
```
jhesketh, line 184: If this is something we expect to always work having some kind of limit and an exception could be helpful

src/gravel/controllers/kv.py
```diff
@@ -12,125 +12,395 @@
...
+
+        logger.debug("Shutting down cluster connection")
+        if self._config_watch:
+            self._config_watch.close()
+            # Need to set this to None, so it's deallocated before the
```
jhesketh, line 189: This might be a python librados bug where `.close()` should actually clean things up (but is reasonable for here)

src/gravel/controllers/kv.py
```diff
@@ -12,125 +12,395 @@
...
+                    values = list(self._db[k] for k in keys)
+                    if keys and not has_aquarium_pool:
+                        with rados.WriteOpCtx() as op:
+                            logger.info(
+                                f"Pushing {keys} to newly created aquarium pool"
+                            )
+                            self._ioctx.set_omap(op, keys, values)  # type: ignore
+                            self._ioctx.operate_write_op(op, "kvstore")
+                    # TODO: need timeouts, error handlers etc on watch
+                    self._config_watch = self._ioctx.watch(
```
jhesketh, line 163: (the ioctx object seems to be really lacking any upstream documentation that I can find :-()

src/gravel/controllers/kv.py
jhesketh: Perhaps those need to be tracked in a separate cache so not to confuse ourselves with the state of the cluster?

But I do see the below comments on getting back into sync being very difficult.

src/gravel/controllers/kv.py
```diff
@@ -12,125 +12,395 @@
...
+                    #     _write_complete,
+                    #     _write_safe)
+                    # Q: can we determine upfront whether a write will block?
+                    self._ioctx.operate_write_op(op, "kvstore")
+                    # This next notifies all watchers *INCLUDING* me!
+                    self._ioctx.notify("kvstore", key)
+            except Exception as e:
+                # TODO: deal with this (log it? ignore it?)
+                # e.g. RADOS rados state (You cannot perform that operation on a Rados object in state configuring.)
+                logger.error(str(e))
```
jhesketh, line 283: (nit) if you use `logger.exception` it includes the stacktrace

src/gravel/controllers/kv.py
```diff
@@ -12,125 +12,395 @@
...
-        assert self._client
+        # This is ugly, because pulling it out of omap is nice (there's a prefix
+        # arg), but getting it from dbm means iterating through everything.
+        # On the plus side, we're probably not talking about a lot of data...
+        # But there's a perf optimization argument here for taking the values from
+        # omap, setting them in the dbm, and returning the values from *omap* if
+        # present, then only falling back to iterating dbm if we don't have the
+        # values, rather than using the simpler logic in get().  But that might be
+        # premature optimization, so possibly more straightforward to do the
+        # simple implementation, then comment this for later use.
```
jhesketh, line 341: Starting with simple seems very reasonable to me.

src/gravel/controllers/kv.py
```diff
@@ -12,125 +12,395 @@
...
         if not value:
             return None
         return value.decode("utf-8")
 
+    async def get(self, key: str) -> Optional[str]:
+        """Get value for provided key"""
+        # Wraps _get() so we can use _get() outside asyncio in _config_notify()
+        return self._get(key)
+
     async def get_prefix(self, key: str) -> List[str]:
```
jhesketh, line 331: (nit the param here should probably be `prefix` not `key` (or perhaps `key_prefix`))

src/gravel/controllers/kv.py
```diff
@@ -12,125 +12,395 @@
...
+        # omap, setting them in the dbm, and returning the values from *omap* if
+        # present, then only falling back to iterating dbm if we don't have the
+        # values, rather than using the simpler logic in get().  But that might be
+        # premature optimization, so possibly more straightforward to do the
+        # simple implementation, then comment this for later use.
         values = []
-        async for value, _ in self._client.get_prefix(key):  # type: ignore[attr-defined]
-            values.append(value.decode("utf-8"))
+
+        if self._ioctx:
```
jhesketh, line 344: (nit) this could possibly be pulled into a "refresh cache" method.

src/gravel/tests/conftest.py
```diff
@@ -311,6 +286,11 @@ class FakeStorage(Storage):  # type: ignore
             total = 2000  # type: ignore
 
         gstate: GlobalState = GlobalState()
+        # gstate's KV store will start a thread to try to connect to the
```
jhesketh, line 289: A better fix here is to fake the connection method.

jhesketh (commented):
Adding my review now as I've been sitting on this for a few days.

In general I like this approach but I still need to give some more thought as to the race conditions of the cache and initial bootstrapping.

I want to get this running on a local cluster to see how it works but have been held up.

src/gravel/tests/conftest.py
tserong: How do I do that? when gstate is constructed it creates a real KV (which starts the connector thread), then later we replace that with a FakeKV (`gstate._kvstore = FakeKV()`). I could fake the connection method in FakeKV, but by then it's too late.

src/gravel/controllers/kv.py
```diff
@@ -12,125 +12,395 @@
...
-        self._is_closing = False
+        if "rados" not in sys.modules:
+            raise MissingSystemDependency("python3-rados module not found")
+        # TODO: this should be created somewhere else (src/gravel/controllers/config.py?)
+        var_lib_aquarium: Path = Path("/var/lib/aquarium")
+        if not var_lib_aquarium.exists():
+            var_lib_aquarium.mkdir(0o700)
+        # this will fail with "_gdbm.error: [Errno 11] Resource temporarily unavailable: '/var/lib/aquarium/kvstore'
+        # if someone else has it open ... need KV to be a singleton?
+        # Or, do we only open it when reading/writing in individual methods?
```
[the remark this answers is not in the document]
tserong: Practically, we probably gain nothing from not keeping it open. I did spend a while experimenting with a couple different ways of opening/closing in individual R/W methods, but all that did was make the implementation irritatingly complex. The only reason this came up at all was when I was doing some manual testing while aquarium was running, I wanted to see what was in the cache, and discovered I couldn't use `gdbmtool`, or python interactively, to poke around because the file was locked.

src/gravel/tests/conftest.py
tserong: Oooh, I just had a thought - if I change the GlobalState constructor to take a KV as a param (which defaults to KV) then I should be able to override that in the test with a FakeKV.

src/gravel/controllers/kv.py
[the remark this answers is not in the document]
tserong: I've updated the comment appropriately

src/gravel/controllers/kv.py
```diff
@@ -12,125 +12,395 @@
...
+                    self._ioctx = self._cluster.open_ioctx("aquarium")
+                    # This actually seems to be safe (doesn't trash existing omap
+                    # data if present, which is neat)
+                    self._ioctx.write_full(
+                        "kvstore",
+                        "# aquarium kv store is in this object's omap\n".encode(
+                            "utf-8"
+                        ),
+                    )
+                    # At this point, iff it's a new pool, new object, etc.
```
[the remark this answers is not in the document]
tserong: Fixed

src/gravel/controllers/kv.py
tserong: I've changed it to key_prefix

src/gravel/controllers/kv.py
```diff
@@ -12,125 +12,397 @@
...
+
+        if self._ioctx:
+            try:
+                with rados.ReadOpCtx() as op:
+                    omap_iter, ret = self._ioctx.get_omap_vals(
+                        op,
+                        start_after="",
+                        filter_prefix=key_prefix,
+                        max_return=1000,
+                    )
```
jecluis, line 351: Have you run this through black? I have a feeling this needs to be reformatted.

jecluis (approved):
This looks pretty good. Just run it through black to ensure the formatting and we should be good to go.

src/gravel/controllers/kv.py
jhesketh: Black is passing according to github checks.

tserong: Black actually told me to do that :-)

tserong:
> This looks pretty good. Just run it through black to ensure the formatting and we should be good to go.

I still have some review comments I want to address, but I'm happy if you want to merge this as-is. I can pick those up in a separate PR or PRs.

src/gravel/controllers/kv.py
```diff
@@ -12,125 +12,395 @@
...
+            var_lib_aquarium.mkdir(0o700)
+        # this will fail with "_gdbm.error: [Errno 11] Resource temporarily unavailable: '/var/lib/aquarium/kvstore'
+        # if someone else has it open ... need KV to be a singleton?
+        # Or, do we only open it when reading/writing in individual methods?
+        # That would probably be better... (but would still have problems due to
+        # simultaneous access, but we can always raise an exception in that case I guess?)
+        try:
+            self._db = dbm.open(f"{var_lib_aquarium}/kvstore", "c")
+        except Exception as e:
+            # Only raise exceptions opening db file when *not* running unit tests
```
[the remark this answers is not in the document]
tserong: This is fixed by 3eeb8ef

tserong:
Sorry, just force pushed a tiny tweak, which dismissed your approval @jecluis ...

jecluis (dismissed):

src/gravel/controllers/kv.py
tserong: It's pretty sketchy... I've been bouncing between https://docs.ceph.com/en/latest/rados/api/python/, with the C API as a backup https://docs.ceph.com/en/latest/rados/api/librados/. Also it turns out https://github.com/ceph/ceph/blob/master/src/test/pybind/test_rados.py is a useful source of examples.

src/gravel/controllers/kv.py
tserong: Yep, that works :-)

src/gravel/controllers/kv.py
```diff
@@ -12,125 +12,395 @@
...
+        self._ioctx: Optional[rados.Ioctx] = None
+        # Watches are setup by calls to watch(); this is a hash of keys to
+        # watch IDs and callbacks, e.g.:
+        # {
+        #     "foo": { 1: callback_fn, 2: another_callback_fn },
+        #     "bar": { 3: yet_another_callback }
+        # }
+        # This structure makes it trivial to invoke all registered callbacks
+        # on a given key, but makes it somewhat irritating to cancel a watch
+        # (you have to iterate through all keys to find the watch id to delete)
```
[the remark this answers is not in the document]
tserong: I've added comments to that effect

tserong:
...and there I go dismissing @jecluis's review again by adding more commits :-/ but at least they shouldn't be too controversial.

I still have some fiddling to do with exception raising / error logging and handling cache updates via watches per the remaining review comments, but I can do those in a separate PR if you'd rather get this in as-is.

src/gravel/controllers/kv.py
```diff
@@ -12,125 +12,395 @@
...
+            await asyncio.sleep(1)
+            tries += 1
+        if self._cluster:
+            logger.info(
+                f"ensure_connection: cluster state '{self._cluster.state}' after {tries} tries"
+            )
+        else:
+            logger.info(
+                f"ensure_connection: no cluster exists yet after {tries} tries"
+            )
```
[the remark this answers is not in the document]
tserong: Done in 44327eb

src/gravel/controllers/kv.py
tserong: I think I'm using the watch in a way that isn't usually tested, i.e. it's not in a `with` block, and I'm actually explicitly calling `Rados.shutdown()` elsewhere. The offending `__dealloc__` code is https://github.com/ceph/ceph/blob/a9a4cc9abf201eca9c11a0bf82eedd14f2343fbb/src/pybind/rados/rados.pyx#L2027-L2029. The fix in rados (to avoid having to set the object to `None` after calling `Watch.close()`) is probably this:

```diff
--- a/src/pybind/rados/rados.pyx
+++ b/src/pybind/rados/rados.pyx
@@ -2025,6 +2025,8 @@ cdef class Watch(object):
         return False
 
     def __dealloc__(self):
+        if self.id == 0:
+            return
         self.ioctx.rados.require_state("connected")
         self.close()
 ```

Of course, now that I think of it, it would be enough for me to set the object to None, and _not_ call close, then I'd get the automatic close-on-dealloc. But, I _still_ think it's a bug in librados that an already-explicitly-closed watch raises an error when it goes out of scope after the cluster shuts down, so I've made a note to do something about addressing that upstream :-)

tserong: ...and here's the proposed librados fix: https://github.com/ceph/ceph/pull/43107

src/gravel/controllers/kv.py
tserong: Done!

tserong:
jenkins run leap

tserong:
OK, I am totally done with this PR now. Someone (@jecluis @jhesketh) please approve and merge it :-)

There are four items of followup work, which IMO can happen in separate PRs:

1) Spend some more time looking at cache semantics / updates on watches / refresh cache on startup.
2) Figure out whether or not we actually need locks around usage of `self._ioctx` and `self._cluster`.
3) Check exactly what happens if the cluster dies horribly while aquarium is running, and if the result is bad, fix it.
4) Unit tests for the KV class (I held off on this, because depending on the above three items, I'd have had to rewrite them anyway).

jecluis:
@tserong happy to merge once the conflict if fixed. :)

jecluis (approved):

PR #637, opened 2021-09-01
[RFC] tools: Add pre-commit config

irq0:
Add configuration for https://pre-commit.com - "A framework for
managing and maintaining multi-language pre-commit hooks."

Hooks:

black and isort like we already have in our tox
configuration

flake8 - "Your Tool For Style Guide Enforcement"

tailing-whitespace - Removes tailing whitespace
check-yaml - Simple yaml checker. Mostly for the config itself
check-added-large-files - Prevents adding large files
end-of-file-fixer - Always have a newline at the end of a file
check-docstring-first - Code must come after a docstring
debug-statements - e.g python debugger imports
requirements-txt-fixer - Sort requirements.txt alphabetically

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+max-line-length = 80
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,24 @@
+repos:
+-   repo: https://github.com/ambv/black
+    rev: 21.7b0
+    hooks:
+    - id: black
+      language_version: python3.8
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.0.1
+    hooks:
+    - id: trailing-whitespace
+    - id: check-yaml
+    - id: check-added-large-files
+    - id: end-of-file-fixer
+    - id: check-docstring-first
+    - id: debug-statements
+    - id: requirements-txt-fixer
+-   repo: https://gitlab.com/pycqa/flake8
+    rev: 3.9.2
+    hooks:
+    - id: flake8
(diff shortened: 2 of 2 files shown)


irq0:
Please note, this only adds the config, but doesn't apply it to the whole repo (pre-commit run --all-files).

irq0:
Do we want something like this? 

Other hooks that we might consider:
- mypy
- pyupgrade - "A tool (and pre-commit hook) to automatically upgrade syntax for newer versions of the language."

mgfritch (commented):
> Do we want something like this?
> 

this is an interesting tool!

> Other hooks that we might consider:
> 
> * mypy
> * pyupgrade - "A tool (and pre-commit hook) to automatically upgrade syntax for newer versions of the language."

we are currently using mypy via tox:
https://github.com/aquarist-labs/aquarium/blob/497e98b3e6a41b5c1e2244f987190ba152780d00/src/tox.ini#L36


rather than duplicate in two places would it be worth calling tox from pre-commit?
similar to this:
https://github.com/aquarist-labs/aquarium/blob/497e98b3e6a41b5c1e2244f987190ba152780d00/.github/workflows/tox.yml#L24

.flake8
```diff
@@ -0,0 +1,2 @@
+[flake8]
```
jhesketh, line 1: We don't need flake8 if using black.

jhesketh (changes requested):
We should use `tox`[0] to manage our tests and linting environment so that it is consistent with CI and also takes care of python dependencies correctly (ie this should replace flake8, black, isort, and also do our unit tests with pytest).

[0] https://pre-commit.com/#usage-with-tox

PR #648, opened 2021-09-13
[tool] Update kiwi and box version requirement to update the build

AvengerMoJo:
As title, kiwi update a new release the old version build images are no longer exist. So I need to update the version number in order to allow the build to work for all the platforms.  

Signed-off-by: Alex Lau (AvengerMoJo) <alau@suse.com>


kshtsk (approved):
lgtm

PR #642, opened 2021-09-03
[DNM][WIP] tools/aqua: multiple deployment types

kshtsk:
Signed-off-by: Kyr Shatskyy <kyrylo.shatskyy@suse.com>

## Checklist
- [ ] References issues, create if required
- [ ] Updates documentation if necessary
- [ ] Includes tests for new functionality or reproducer for bug

---

<details>
<summary>Show available Jenkins commands</summary>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>

PR #433, opened 2021-04-16
run_aquarium.sh: default to running with system deps

tserong:
On the assumption that run_aquarium.sh is usually run inside the MicroOS image (and not on the dev host), let's default to using
system deps. This commit removes the --with-system-deps option and adds a --use-venv option for cases when using the venv is
desirable.

Signed-off-by: Tim Serong <tserong@suse.com>

## Checklist
- [ ] References issues, create if required
- [ ] Updates documentation if necessary
- [ ] Includes tests for new functionality or reproducer for bug

---

<details>
<summary>Show available Jenkins commands</summary>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>

PR #650, opened 2021-09-14
deps: Fix potential dependency issues on dev machines

jhesketh:
Because setup-dev.sh is ran on the developers machine, it is not okay
to use the system packages which may conflict with the requirements.txt
aquarium is trying to install (for example, the system packages may
already supply a dependency which will mean that pip will then not
install it into the venv).

Instead we should ensure the venv is created without system packages.
This is what is eventually copied into the virtual machine. Before we
copy it in though we need to re-enable the system packages for
librados to work, so we do that manually with sed.

### DIFF
--- a/tools/requirements.txt
+++ b/tools/requirements.txt
@@ -1,0 +1,5 @@
+aiohttp
+click==7.1.2
+pydantic==1.7.3
+kiwi==9.23.56
+kiwi-boxed-plugin==0.2.15
--- a/tools/setup-dev.sh
+++ b/tools/setup-dev.sh
@@ -2,6 +2,7 @@
 
 dependencies_opensuse=(
   "btrfsprogs"
+  "gcc"
   "git"
   "kpartx"
   "make"
@@ -362,17 +363,24 @@ if [ -d venv ] ; then
     echo
 fi
 
-
 # We need system site packages because librados python bindings are only
 # available as a package, they're not on pypi (and Tim thinks they probably
-# never will be). Setting --system-site-packages here means the venv will
-# work properly if someone uses it inside the image, where we need to be
-# able to fall back to system packages to get librados.
-$PYTHON -m venv --system-site-packages venv || exit 1
+# never will be).
+# However, if we set --system-site-packages on creation, the required
+# dependencies may not be installed properly if they already exist in the
+# developers environment.
+# Instead, first create an isolated venv, install the requirements, then
+# enable system packages for future venv activations.
+# This will ensure if someone uses it inside the image, they will be able
+# to fall back to system packages to get librados.
+$PYTHON -m venv venv || exit 1
 
 source venv/bin/activate
 pip install -r src/requirements.txt -U || exit 1
 deactivate
+
+# Modify pyenv.cfg to enable system packages from now onwards.
+sed -i 's/include-system-site-packages = false/include-system-site-packages = true/g' venv/pyvenv.cfg
 
 pushd src/glass &>/dev/null
 npm ci || exit 1
--- a/tools/setup-dev.sh
+++ b/tools/setup-dev.sh
@@ -353,34 +353,24 @@ git submodule update --init || exit 1
 [[ ! -e "src/gravel/cephadm/cephadm.bin" ]] && \
   ln -fs ../ceph.git/src/cephadm/cephadm src/gravel/cephadm/cephadm.bin
 
-if [ -d venv ] ; then
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+
+if [ -d $SCRIPT_DIR/venv ] ; then
     echo
     echo "Detected an existing virtual environment:"
-    echo "  > $(realpath venv)"
+    echo "  > $(realpath $SCRIPT_DIR/venv)"
     if yes_no "Blow it away"; then
-        rm -rf venv || exit $?
+        rm -rf $SCRIPT_DIR/venv || exit $?
     fi
     echo
 fi
 
-# We need system site packages because librados python bindings are only
-# available as a package, they're not on pypi (and Tim thinks they probably
-# never will be).
-# However, if we set --system-site-packages on creation, the required
-# dependencies may not be installed properly if they already exist in the
-# developers environment.
-# Instead, first create an isolated venv, install the requirements, then
-# enable system packages for future venv activations.
-# This will ensure if someone uses it inside the image, they will be able
-# to fall back to system packages to get librados.
-$PYTHON -m venv venv || exit 1
+# Set up Virtual Env for dev tools
+$PYTHON -m venv $SCRIPT_DIR/venv || exit 1
 
-source venv/bin/activate
-pip install -r src/requirements.txt -U || exit 1
+source $SCRIPT_DIR/venv/bin/activate
+pip install -r $SCRIPT_DIR/requirements.txt -U || exit 1
 deactivate
-
-# Modify pyenv.cfg to enable system packages from now onwards.
-sed -i 's/include-system-site-packages = false/include-system-site-packages = true/g' venv/pyvenv.cfg
 
 pushd src/glass &>/dev/null
 npm ci || exit 1
--- a/tools/setup-dev.sh
+++ b/tools/setup-dev.sh
@@ -1,12 +1,13 @@
 #!/bin/bash
 
-dependencies_opensuse_tumbleweed=(
+dependencies_opensuse=(
   "btrfsprogs"
   "git"
   "kpartx"
   "make"
   "python3"
   "python3-kiwi"
+  "python3-devel"
   "nodejs-common"
   "npm"
   "vagrant"
@@ -78,7 +79,7 @@ dependencies_ubuntu=(
   "libvirt-daemon-system"
 )
 
-dependencies_build_python_leap=(
+dependencies_build_python_opensuse=(
   "gcc"
   "automake"
   "bzip2"
@@ -194,7 +195,7 @@ if ${show_dependencies} ; then
 
   case $osid in
     opensuse-tumbleweed | opensuse-leap)
-      echo "  > ${dependencies_opensuse_tumbleweed[*]}"
+      echo "  > ${dependencies_opensuse[*]}"
       ;;
     debian)
       echo "  > ${dependencies_debian[*]}"
@@ -205,7 +206,7 @@ if ${show_dependencies} ; then
     *)
       echo "error: unsupported distribution"
       echo "These are the packages you might need:"
-      echo "  > ${dependencies_opensuse_tumbleweed[*]}"
+      echo "  > ${dependencies_opensuse[*]}"
       exit 1
       ;;
   esac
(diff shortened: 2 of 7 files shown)


tserong (commented):
LGTM, but the Tumbleweed build is now broken, which I suspect is because without system deps, kiwi(?)'s dependency on python3-xattr results in that module being built in the venv, and gcc isn't installed.

This will probably fix it (untested):

```diff --git a/tools/setup-dev.sh b/tools/setup-dev.sh
index 06a33ac..1cd60b2 100755
--- a/tools/setup-dev.sh
+++ b/tools/setup-dev.sh
@@ -2,6 +2,7 @@
 
 dependencies_opensuse_tumbleweed=(
   "btrfsprogs"
+  "gcc"
   "git"
   "kpartx"
   "make"
@@ -79,7 +80,6 @@ dependencies_ubuntu=(
 )
 
 dependencies_build_python_leap=(
-  "gcc"
   "automake"
   "bzip2"
   "libbz2-devel"
```
(The names of the lists of variables above are a bit misleading -  AIUI `dependencies_opensuse_tumbleweed` will always be installed on both tumbleweed _and_ leap, whereas dependencies_build_python_leap is only there on leap)

jhesketh:
Ah, thanks for that. I'll give it a try (but don't have a clean tumbleweed to check so will experiment here)

tserong:
Looks like the tumbleweed build has been stuck for the last couple of hours doing a `git fetch --tags --progress https://github.com/aquarist-labs/aquarium '+refs/heads/*:refs/remotes/origin/*'`...? I wonder if I ask it to run tests again if it'll sort itself out...

tserong:
jenkins run tumbleweed

tserong:
...and down the rat hole we go: `xattr.c:25:10: fatal error: Python.h: No such file or directory`, which I guess means we also need python3-devel installed? But why aren't we hitting this on the leap build?

jhesketh:
Hmm, I wonder if pyenv includes devel files. It'd make sense for it to. Nonetheless I'll try adding it to the system requirements.

tserong (approved):
\o/

kshtsk:
Mixing the host and guest venv is not a good idea per se.
Seems like this PR will conflict with my patch using system libvirt package for the development environment, it needs to double check.

kshtsk:
> Hmm, I wonder if pyenv includes devel files. It'd make sense for it to. Nonetheless I'll try adding it to the system requirements.

exactly, pyenv is building a python and already includes dev headers.

jhesketh:
> Mixing the host and guest venv is not a good idea per se.
> Seems like this PR will conflict with my patch using system libvirt package for the development environment, it needs to double check.

Actually we're mixing the system deps and venv less with this patch. Previously the venv was created with `--system-site-packages`. This patch avoids that and installs everything inside the venv. We only re-enable it for when aquarium itself uses the venv inside the virtual machine.

Which patch are you worried this conflicts with?

kshtsk:
The #642

jhesketh:
So is the issue that aqua needs `libvirt-python` which requires site-packages to work?

If `libvirt-python` can be installed without site-packages then there is no harm in this patch here.

If it can't, then an alternative may be for a different venv for aqua than for aquarium itself.

This PR fixes a real problem that I encountered. Without recreating the specifics (since it was a while ago), it's obvious that there are potential incompatibilities, version conflicts, and other dependencies problems that we may see where we are using this venv in a developers environment and then in the actual deployed image itself.

kshtsk:
> So is the issue that aqua needs `libvirt-python` which requires site-packages to work?
> 
Not exactly and not only. Indeed aqua is using site-packages because my patch was taking into account this, once this merged I will need to update it to build Libvirt-python for tumbleweed, for example.

> If `libvirt-python` can be installed without site-packages then there is no harm in this patch here.

Yes it can but we have more fundamental issue, we should not use host system virtualenv on guest system, because  in general not only files layout or packages differed but also architecture. Also, it (libvirt-python) is  not needed on guest system.

> 
> If it can't, then an alternative may be for a different venv for aqua than for aquarium  

According to docs we use venv (which is produced setup-dev.sh, i.e. development environment, in comparison production like) for aqua to spin up the cluster. I thought @tserong was working on splitting guest venv from host one.

> This PR fixes a real problem that I encountered. Without recreating the specifics (since it was a while ago), it's obvious that there are potential incompatibilities, version conflicts, and other dependencies problems that we may see where we are using this venv in a developers environment and then in the actual deployed image itself.

I believe it fixes some of your real issues, but as I said above it will not fix them if we continue share host venv with guest one. If this something what is blocking you I guess you can merge it. But if not, then probably we had to look for better solution?

tserong:
> I thought @tserong was working on splitting guest venv from host one.

There is no venv used inside the guest in normal operation. Unless I'm forgetting something, the only way to have a venv in there is if you're invoking `run_aquarium.sh --use-venv` manually (see https://github.com/aquarist-labs/aquarium/pull/433).

jecluis:
I'm confused about the premise of this PR, to be honest. It does feel like the host venv is being passed to, and used by, the guest system, which is something that should not happen under normal circumstances. All of the dependencies should be provided by the image itself. There may be times when one is trying out new dependencies that one might require a venv, but it's preferable to init a new venv dedicated to the guest, from the guest, rather than using the one shared by the host.

tserong:
I approved this on the assumption it was fixing something for Josh on the host, not in the guest. Maybe I was mistaken?

jecluis:
Yeah, I think that needs to be clarified. @jhesketh ?

jhesketh:
So the issue is that `setup-dev.sh` is ran on the developers machine. This script creates the virtual environment for Aquarium and installs the dependencies into that.

Vagrant then later mounts the aquarium folder into /srv/aquarium, where the venv is consumed (or at least the way I was using it it was).

The problem here is that because the venv was created with `--system-site-packages` on the devs machine, then the venv when ran in the libvirt guest is not complete.

Regardless, we should really have all the dependencies in a virtualenv, rather than on the system. The only reason for using the system is because we can't fulfill one of those deps. So this creates a workaround whereby most of the packages are in the venv.

Hope that makes sense? Having said all that, perhaps I'm missing something and my dev env is just in a really weird state.

kshtsk:
> So the issue is that `setup-dev.sh` is ran on the developers machine. This script creates the virtual environment for Aquarium and installs the dependencies into that.
> 
> Vagrant then later mounts the aquarium folder into /srv/aquarium, where the venv is consumed (or at least the way I was using it it was).
> 
> The problem here is that because the venv was created with `--system-site-packages` on the devs machine, then the venv when ran in the libvirt guest is not complete.
> 

Just do not use the `venv` in libvirt guest, because it for the host only.

> Regardless, we should really have all the dependencies in a virtualenv, rather than on the system. The only reason for using the system is because we can't fulfill one of those deps. So this creates a workaround whereby most of the packages are in the venv.
> 
> Hope that makes sense? Having said all that, perhaps I'm missing something and my dev env is just in a really weird state.

jhesketh:
So I've given this some more thought and I think the issue here is that we are using on venv for multiple purposes.

Firstly, in general, we should not use site-packages because they can cause conflicts and unexpected versions. Instead we should try to install all of the packages inside of a venv.

In the case of `setup-dev.sh`, it needs a venv for running what is inside of `tools/`. It doesn't necessarily need the same `requirements.txt` that gravel does. For that reason, I think it should have its own requirements file that is installed. This also helps #642 whereby `libvirt-python` and `pytest` can be added to it (and as a separate discussion on that PR we can investigate if using site-packages makes more sense due to the nature of libvirt).

Then, for development, the `run_aquarium.py --use-venv` script should use a different venv that is created within the guest using gravel's requirements.txt. It can also take care of using site packages if it needs to.

I'll rework this and see if it makes more sense.

tools/requirements.txt
```diff
@@ -0,0 +1,5 @@
+aiohttp
+click==7.1.2
+pydantic==1.7.3
+kiwi
+kiwi-boxed-plugin
```
tserong, line 5: These two should specify versions (since 4283653):
```
kiwi==9.23.56
kiwi-boxed-plugin==0.2.15
```
Also, can they now be removed from src/requirements.txt?

tserong (commented):
LGTM with one tweak. What do you think @kshtsk?

Note that the leap and tumbleweed build failures seem unrelated (looks like the "dive into aquarium" thing has been broken for a few days, I assume due to the migration from angular material to bootstrap)

kshtsk:
jenkins run leap

kshtsk:
jenkins run leap

tools/requirements.txt
jhesketh: Yep, good catches, thanks.

kshtsk:
jenkins test opensuse

jecluis (changes requested):
I'm pretty sure this needs to be documented, as now we're having different venvs and the FHTH guide does not account for that.

jhesketh:
> I'm pretty sure this needs to be documented, as now we're having different venvs and the FHTH guide does not account for that.

This PR updates the FZTH guide to use the correct new venv. I don't think we need to explain it anywhere else, as there isn't really a venv used for aquarium (just the tools) in the usual developer workflow.

jecluis:
> > I'm pretty sure this needs to be documented, as now we're having different venvs and the FHTH guide does not account for that.
> 
> This PR updates the FZTH guide to use the correct new venv. I don't think we need to explain it anywhere else, as there isn't really a venv used for aquarium (just the tools) in the usual developer workflow.

A developer workflow that, if not documented, will become tribal knowledge.

jhesketh:
> > > I'm pretty sure this needs to be documented, as now we're having different venvs and the FHTH guide does not account for that.
> > 
> > 
> > This PR updates the FZTH guide to use the correct new venv. I don't think we need to explain it anywhere else, as there isn't really a venv used for aquarium (just the tools) in the usual developer workflow.
> 
> A developer workflow that, if not documented, will become tribal knowledge.

It is documented though. Nothing has changed for the developer here unless I'm missing something?

tserong:
> > > > I'm pretty sure this needs to be documented, as now we're having different venvs and the FHTH guide does not account for that.
> > > 
> > > 
> > > This PR updates the FZTH guide to use the correct new venv. I don't think we need to explain it anywhere else, as there isn't really a venv used for aquarium (just the tools) in the usual developer workflow.
> > 
> > 
> > A developer workflow that, if not documented, will become tribal knowledge.
> 
> It is documented though. Nothing has changed for the developer here unless I'm missing something?

Joao is hopefully asleep :-) but your FZTH change looks fine to me.

The only thing I can think of is that maybe he's suggesting we should explicitly explain the runtime venv which doesn't usually exist, but which is created if you decide to invoke `./tools/run_aquarium.sh --use-venv`, and that that's different from the tools venv?

jhesketh:
> > > > > I'm pretty sure this needs to be documented, as now we're having different venvs and the FHTH guide does not account for that.
> > > > 
> > > > 
> > > > This PR updates the FZTH guide to use the correct new venv. I don't think we need to explain it anywhere else, as there isn't really a venv used for aquarium (just the tools) in the usual developer workflow.
> > > 
> > > 
> > > A developer workflow that, if not documented, will become tribal knowledge.
> > 
> > 
> > It is documented though. Nothing has changed for the developer here unless I'm missing something?
> 
> Joao is hopefully asleep :-) but your FZTH change looks fine to me.
> 
> The only thing I can think of is that maybe he's suggesting we should explicitly explain the runtime venv which doesn't usually exist, but which is created if you decide to invoke `./tools/run_aquarium.sh --use-venv`, and that that's different from the tools venv?

I've added a bit of extra explanation that should hopefully cover this.

jecluis (approved):

tools/setup-dev.sh
```diff
@@ -351,26 +353,23 @@ git submodule update --init || exit 1
...
-# available as a package, they're not on pypi (and Tim thinks they probably
-# never will be). Setting --system-site-packages here means the venv will
-# work properly if someone uses it inside the image, where we need to be
-# able to fall back to system packages to get librados.
-$PYTHON -m venv --system-site-packages venv || exit 1
-
-source venv/bin/activate
-pip install -r src/requirements.txt -U || exit 1
+source $SCRIPT_DIR/venv/bin/activate
+pip install -r $SCRIPT_DIR/requirements.txt -U || exit 1
```
kshtsk, line 372: @jhesketh though the PR is merged, I don't understand why do you want this?????
Once get rid off --system-site-packages for tools/venv I now it will rebuild libvirt-python instead of using the system one if we put it to requirements.txt.

PR #665, opened 2021-09-20
hardware: Validate individual node requiements

jhesketh:
This is a very crude validation of the localhosts hardware requirements.
The frontend can check that the node passes validation before allowing
it to be bootstrapped or enrolled into the cluster.

No ongoing monitoring occurs, and we could still allow a user to proceed
to bootstrap if they wanted to.


jecluis (changes requested):
generally looks good, but I really think the frontend will benefit from error codes (even if an error msg is associated with it) a lot more than free form strings.

src/gravel/controllers/nodes/local.py
```diff
@@ -0,0 +1,159 @@
...
+    actual_disk: int = Field("Actual size of root disk (GB)")
+    error: str = Field("Root disk didn't meet requirements")
+
+
+class LocalhostQualifiedModel(BaseModel):
+    qualified: bool = Field("The localhost passes validation")
+    errors: List[str] = []
+
+
+async def validate_cpu() -> CPUQualifiedModel:
```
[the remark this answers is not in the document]
jhesketh: There is a potential future optimisation where some of these checks might be IO or network bound (and maybe psutil will add async support in the future). I'll update the checks to await the tasks asynchronously and keep these for consistency.

src/gravel/controllers/nodes/local.py
```diff
@@ -0,0 +1,159 @@
...
+        errors.append(qualified_mem.error)
+
+    qualified_disk: RootDiskQualifiedModel = await validate_root_disk()
+    if not qualified_disk.qualified:
+        qualified = False
+        errors.append(qualified_disk.error)
+
+    result = LocalhostQualifiedModel(
+        qualified=qualified,
+        errors=errors,
```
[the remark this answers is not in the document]
jhesketh: Yep, good point, thanks.

src/gravel/tests/functional/test_app.py
```diff
@@ -11,19 +11,226 @@
...
+    # Everything passes
+    mocker.patch("psutil.cpu_count", return_value=8)
+    mocker.patch(
+        "psutil.virtual_memory", return_value=FakeMemory(total=33498816512)
+    )
+    mocker.patch("psutil.disk_usage", return_value=FakeDisk(total=20000000000))
+
+    response = await async_client.get("/api/local/status")
+    assert response.status_code == 200
+    response_json = json.loads(response.text)
```
jecluis, line 76: why not parsing the model?

src/gravel/tests/functional/test_app.py
```diff
@@ -11,19 +11,226 @@
...
+    )
+
     import asyncio
 
     await asyncio.sleep(5)
 
     response = await async_client.get("/api/local/status")
     assert response.status_code == 200
-    assert '{"inited":true,"node_stage":0}' == response.text
+    assert (
```
jecluis, line 49: this seems prone to error. Better use the model and parse the raw text IMO.

src/gravel/controllers/nodes/local.py
```diff
@@ -0,0 +1,185 @@
...
+    if not (
+        cpu_qualified.qualified
+        and mem_qualified.qualified
+        and root_disk_qualified.qualified
+    ):
+        all_qualified = False
+
+    result = LocalhostQualifiedModel(
+        all_qualified=all_qualified,
+        errors=errors,
```
jecluis, line 180: this is an empty list?

src/gravel/controllers/nodes/local.py
```diff
@@ -0,0 +1,185 @@
...
+    error: str = ""
+    status: HardwareQualifiedEnum = HardwareQualifiedEnum.QUALIFIED
+
+    if actual_cpu < min_cpu:
+        qualified = False
+        error = (
+            "The node does not have a sufficient number of CPU cores. "
+            "Required: %d, Actual: %d." % (min_cpu, actual_cpu)
+        )
+        status = HardwareQualifiedEnum.ERROR
```
jecluis, line 86: The problem with this is that it is not conveying what the error is to the consumer. As we discussed before, the frontend will not be using the error message as is. What it will need is an error code and information that allows the frontend to, depending on the error code, provide insight to the user.

I'd suggest having error enums for different scenarios, even if at the moment we only support `some_value < target_value`.

src/gravel/controllers/nodes/local.py
```diff
@@ -0,0 +1,185 @@
...
+import psutil
+from pydantic import BaseModel, Field
+
+# Minimum requirements for individual host validation:
+# NOTE(jhesketh): These are obviously hardcoded for the moment, but may be
+#                 better moved into some kind of support matrix (eg feature
+#                 vs hardware etc).
+#                 These are also currently somewhat arbitrary and will need
+#                 tuning.
+AQUARIUM_MIN_CPU = 2
```
jecluis, line 27: I think it would be preferable to use bytes here instead of GB, and let the frontend do the conversion in a consistent manner.

jecluis (changes requested):

src/gravel/controllers/nodes/local.py
```diff
@@ -0,0 +1,185 @@
...
+    all_qualified: bool = True
+    errors: List = []
+
+    cputask = asyncio.create_task(validate_cpu())
+    memtask = asyncio.create_task(validate_memory())
+    disktask = asyncio.create_task(validate_root_disk())
+
+    cpu_qualified: CPUQualifiedModel = await cputask
+    mem_qualified: MemoryQualifiedModel = await memtask
+    root_disk_qualified: RootDiskQualifiedModel = await disktask
```
votdev, line 170: What about using this?
```
cpu_qualified: CPUQualifiedModel, mem_qualified: MemoryQualifiedModel, root_disk_qualified: RootDiskQualifiedModel = await asyncio.gather(validate_cpu(), validate_memory(), validate_root_disk())
```
OK, it will not reduce code lines, but complexity.

jecluis: that's a really good idea

src/gravel/controllers/nodes/local.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+class HardwareQualifiedEnum(int, Enum):
+    QUALIFIED = 0
+    ERROR = 1
+
+
+class CPUQualifiedModel(BaseModel):
+    qualified: bool = Field("The CPU is sufficient")
+    min_cpu: int = Field("Minimum number of CPU's")
```
s0nea, line 39: I'm wondering if it's possible/sensible to use just `min` instead of `min_cpu` and `actual` for `actual_cpu` (same for `min_mem`, `actual_mem`, `min_disk`, `actual_disk`)? That way the entries would be more generic. Or is there an actual reason against it?

src/gravel/tests/functional/test_app.py
jhesketh: Good point, done.

src/gravel/controllers/nodes/local.py
jhesketh: This was leftover from a refactor and should be removed.

src/gravel/controllers/nodes/local.py
jhesketh: Yep, very good point, thanks.

src/gravel/controllers/nodes/local.py
jhesketh: I assume this comment was meant for memory and maybe disk, but good idea.

src/gravel/controllers/nodes/local.py
jhesketh: Yep, agreed.

src/gravel/controllers/nodes/local.py
jhesketh: The main reason is that we may validate more about a piece. For example, we may have `min_cores`, `min_clock` on the CPU and `min_speed`, `min_mem` on the memory. I should update `min_cpu` to be `min_threads` since that is what we are checking.

s0nea: I see, thanks :+1:

jecluis:
DCO is failing because the commit's author email is different from the Signed-off-by email. @jhesketh you'll need to change one of the two, but I'm guessing you'll likely want to reset the author on the individual commits.

```
DCO

You have 7 commits incorrectly signed off. To fix, first ensure you have a local copy of your branch by checking out the pull request locally via command line. Next, head to your local branch and run:

git rebase HEAD~7 --signoff

Now your commits will have your sign off. Next run

git push --force-with-lease origin hardware

Commit sha: 6549b05, Author: Joshua Hesketh, Committer: Joshua Hesketh; Expected "Joshua Hesketh josh@nitrotech.org", but got "Joshua Hesketh jhesketh@suse.com".
Commit sha: f4634c5, Author: Joshua Hesketh, Committer: Joshua Hesketh; Expected "Joshua Hesketh josh@nitrotech.org", but got "Joshua Hesketh jhesketh@suse.com".
Commit sha: 5eca18b, Author: Joshua Hesketh, Committer: Joshua Hesketh; Expected "Joshua Hesketh josh@nitrotech.org", but got "Joshua Hesketh jhesketh@suse.com".
Commit sha: 04787f8, Author: Joshua Hesketh, Committer: Joshua Hesketh; Expected "Joshua Hesketh josh@nitrotech.org", but got "Joshua Hesketh jhesketh@suse.com".
Commit sha: 300f884, Author: Joshua Hesketh, Committer: Joshua Hesketh; Expected "Joshua Hesketh josh@nitrotech.org", but got "Joshua Hesketh jhesketh@suse.com".
Commit sha: 6a22be1, Author: Joshua Hesketh, Committer: Joshua Hesketh; Expected "Joshua Hesketh josh@nitrotech.org", but got "Joshua Hesketh jhesketh@suse.com".
Commit sha: abfb2c4, Author: Joshua Hesketh, Committer: Joshua Hesketh; Expected "Joshua Hesketh josh@nitrotech.org", but got "Joshua Hesketh jhesketh@suse.com".
```

jecluis (changes requested):
LGTM, but we should change the error message to include units. This will mean changing the tests too. But once that's done, I'm more than happy to give it the green check mark.

PR #681, opened 2021-09-28
glass: Use echarts in system information widget

Devp00l:

echarts_systeminformation

This PR introduces the echarts library that will replace ngx-charts in the future. Only the performance widget has to be converted after this in order to fully replace the use of ngx-charts.

This PR also fixes a strange bug where all widgets got an alt tag with their name that overlaid the widget template.

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>


votdev:
It comes out that this is a flex-layout related issue. Need to adapt the widget to fix the issue.

image
.

PR #689, opened 2021-09-30
images: add comment about man pages not being installed

tserong:
When the image is built, <rpm-excludedocs>true</rpm-excludedocs>
makes sure man pages aren't installed, which is one of those things
you exclude to try to keep image sizes down.  This is the build-time
equivalent of rpm.install.excludedocs in /etc/zypp/zypp.conf.

I'm not suggesting we change this setting, just documenting it so
that the next time someone's poking around and wondering why there's
no man pages, we have the answer.

## Checklist

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>


images/aquarium/config.xml
```diff
@@ -72,6 +72,8 @@
         <packagemanager>zypper</packagemanager>
         <locale>en_US</locale>
         <keytable>us</keytable>
+        <!-- If you're wondering why there's no manpages included in the
+             image, this next setting is what ensures they're not installed -->
```
asettle, line 76: Maybe I'm just daft, but this is really confusing. "If you're wondering why this thing doesn't thing, the next thing is why that thing doesn't thing"

How about, "The thing doesn't work and this is by design. See below"

jecluis: Maybe I'm just daft, but it sounds perfectly reasonable to me xD

jecluis (approved):

jecluis:
@tserong I'll leave it up to you to figure out what the other native speaker means, and whether you want to adjust the wording or not. From my pov, this is good to go.

tserong:
OK, I've changed the wording so it's less "Tim's stream of consciousness when solving a problem" and more "here's the thing" :-)

PR #693, opened 2021-10-05
glass: Improve dashboard widgets

votdev:
- Fix various issues related to dashboard widgets, e.g. translate titles or make sure the widget is rendered correct in error state
- Add events dashboard widget and API endpoint (mocked data right now)

## Checklist

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>


votdev:
> Am I the only one who is getting a vertical scroll bar in the system information widget now? - It's happening to me on Chrome and Firefox.
> 
> Firefox: Screenshot_2021-10-05_16-57-53_firefox
> 
> Chrome: Screenshot_2021-10-05_16-57-57_chrome

No, this is correct as you can see in the `Events` widget above. But it shouldn't be displayed in the `System Information` widget with that height in your case.

s0nea:
> No, this is correct as you can see in the `Events` widget above. But it shouldn't be displayed in the `System Information` widget with that height in your case.

Well yes, that's actually what I mean. It's not that there **is** a scroll bar now. I was wondering why a scroll bar is displayed when it isn't needed :smile:

votdev:
> > No, this is correct as you can see in the `Events` widget above. But it shouldn't be displayed in the `System Information` widget with that height in your case.
> 
> Well yes, that's actually what I mean. It's not that there **is** a scroll bar now. I was wondering why a scroll bar is displayed when it isn't needed smile

I've modified the system information widget a little bit to prevent showing the scroll bar too soon. The scroll bar will be shown when the content overlaps the widget. For the system information widget this is a little bit tricky because the graphs are bigger than this is visible. Because of that the scroll bar is shown earlier than you expect.

Bildschirmfoto von 2021-10-07 12-07-53

PR #694, opened 2021-10-07
aquarium: minimal MicroOS image & custom registry support

jecluis:
This patch set introduces a minimal MicroOS based image.

Because this image does not include containers, by design, we need to obtain these from somewhere. We also introduce mechanisms to obtain containers from specified, configurable sources. Some refactor was required to ensure we don't depend on upstream containers prior to the source container being specified.

Additionally, we're dropping the requirement of a `ceph.git` submodule. Instead, we'll now follow Ceph's `master` branch so we keep up with the next Quincy release. We'll fix the release at a later point in time.

Please note that while this patch set allows configuring container sources, at the moment we are not exposing this to the user. I'm hoping we can merge this before working on the frontend.

At the moment, default image points to `registry.opensuse.org` for a master-based, python 3.8 Ceph image.

### DIFF
--- a/tools/build-image.sh
+++ b/tools/build-image.sh
@@ -28,7 +28,6 @@ options:
   -n NAME | --name NAME     Specify build name (default: aquarium).
   -c | --clean              Cleanup an existing build directory before building.
   -t | --type IMGTYPE       Specify image type (default: vagrant).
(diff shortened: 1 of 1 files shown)


jecluis:
I'll address tox failures in parallel to reviews.

PR #700, opened 2021-10-12
gravel: inventory: log probe times at debug level unless > 30 seconds

tserong:
Previously, every minute, we'd see "inventory -- probing took 5 seconds"
in the log, which seems mostly like unimportant noise to me, so I've
dropped this to debug level, provided the probe doesn't take more than
30 seconds to complete.  If it's taking *that* long, maybe we do care
to investigate further (although the choice of 30 seconds is pretty
arbitrary on my part).

## Checklist

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>

### DIFF
--- a/src/gravel/controllers/resources/inventory.py
+++ b/src/gravel/controllers/resources/inventory.py
@@ -70,7 +70,11 @@ class Inventory(Ticker):
         start: int = int(time.monotonic())
         nodeinfo = await cephadm.get_node_info()
         diff: int = int(time.monotonic()) - start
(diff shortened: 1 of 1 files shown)


votdev:
jenkins test leap

tserong:
ok, looks like jenkins is just stuck/stalling - @jecluis do you mind re-approving this so I can merge it and forget about it? :-)

PR #699, opened 2021-10-08
glass: add network page

s0nea:
Please note: this is just a draft. It does only represent the current status. There are still a lot of things missing:

- The REST API endpoint. Including needed request and response data.
- The `DeclarativeFormComponent` is currently not capable of deactivating/hiding fields depending on other fields values. In other words: it's not possible to deactivate/hide the fields `IP Address`, `Netmask` and `Gateway` if `Type` is set to `DHCP`.
- ... I'll continue the list as soon as I'm aware of additional questions :smile: 

Screenshot_2021-10-08_15-40-56

Screenshot_2021-10-08_15-44-08

Signed-off-by: Tatjana Dehler <tdehler@suse.com>

## Checklist
- [ ] References issues, create if required
- [ ] Updates documentation if necessary
- [ ] Includes tests for new functionality or reproducer for bug

---

<details>
<summary>Show available Jenkins commands</summary>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>

PR #702, opened 2021-10-12
glass: Add constraint support

votdev:
- Use constraints to declare form field state changes
- Use constraints in the requiredIf validator

## Checklist
- [x] Includes tests for new functionality or reproducer for bug

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>


votdev:
> Great, thank you for adding the feature +1 slightly_smiling_face
> 
> There is one thing I noticed while testing this PR as part of #699 . It seems like the `readonly` constraint is not validated at component initialization. For example if the type of a network device is 'DHCP' already, the address fields are not disabled. They are only disabled if I switch the type from 'DHCP' to 'static' and back to 'DHCP'

I'm aware of this issue and was in hope that we can workaround this by setting the fields to the origin state they need; e.g. setting the address field to readonly by default.
I will have a look into the issue to find a proper way to fix that.

PR #711, opened 2021-10-18
.github: drop `jenkins test` command

kshtsk:
There may be a confusion between which command is better to use
`jenkins test` or `jenkins run`, so we better drop one of them
from the docs and hints.

Also adding info about Ubuntu trigger phrase.

## Checklist

---

<details>

- `jenkins test tumbleweed`
- `jenkins run tumbleweed`
- `jenkins test leap`
- `jenkins run leap`

</details>

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -35,9 +35,8 @@ https://github.com/aquarist-labs/aquarium/blob/main/CONTRIBUTING.md
 <details>
 <summary>Show available Jenkins commands</summary>
 
-- `jenkins test tumbleweed`
 - `jenkins run tumbleweed`
-- `jenkins test leap`
 - `jenkins run leap`
+- `jenkins run ubuntu`
 
 </details>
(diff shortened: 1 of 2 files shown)


tserong:
> I want two, can I? :D

That's cute :-) but no. The only time it makes sense to have more than one Signed-off-by line is if a single commit is the work of multiple people, then it's one Signed-off-by line per person.

If you want to sign off with kyrylo.shatskyy@suse.com, you need to set your git config user.email to kyrylo.shatskyy@suse.com, and redo the commit so that that email address is also used as the Author of the commit (something like `git reset HEAD~1 ; git config user.email kyrylo.shatskyy@suse.com ; git commit -s ... `). Or, you can continue to author commits as kyrylo.shatskyy@gmail.com and also use that for the sign off. Whatever. Just so long as the Author of the commit is the same as the Signed-off-by line for single author commits (a correct Signed-off-by line with the same name and email address as the Author should be added to the commit message automatically if you're using `git commit -s`).

jecluis:
@kshtsk I'm okay with merging this once the multiple signed-off-by are fixed

Issue #726, opened 2021-10-22
glass/gravel: Include Devices availability when checking Host Requirements


votdev:
When the assessment reports that the system does not fit the requirements, there is a `Continue anyway (I know what I'm doing)` button to ignore that. But this `ignore this` is not used by the backend because i get the error `No possible deployment solution found.` after pressing the `Install` button.

There are two solutions:
- Remove the `Continue anyway (I know what I'm doing)` button
- Enhance the backend in that way that a `force` property is submitted in the `api/nodes/deployment/start` endpoint which is applied at https://github.com/aquarist-labs/aquarium/blob/a5986f490f3324e22da90d1ff455a8a344264f2a/src/gravel/controllers/nodes/mgr.py#L311.

![Peek 2021-10-22 11-09](https://user-images.githubusercontent.com/1897962/138428319-dbe93cff-380d-4267-ba59-7e5f1e44ff33.gif)

PR #734, opened 2021-10-30
aquarium: check disks requirements

jecluis:
Checks whether we have the right amount of disk devices, with an appropriate size for installation, during host requirements checks.

Fixes: #726


src/aquarium.py
```diff
@@ -24,10 +24,10 @@
 
 from gravel.api import auth, devices, local, nodes, orch, status, users
 from gravel.cephadm.cephadm import Cephadm
+from gravel.controllers.ceph.ceph import Ceph, Mgr, Mon
```
mgfritch, line 27:
```suggestion
from gravel.controllers.ceph import Ceph, Mgr, Mon
```
(nit) feels redundant to type `.ceph.ceph import Ceph`, maybe add these classes to `__init__.py`?

mgfritch (approved):
one nit, otherwise lgtm!

src/aquarium.py
jecluis: Oof. At a later time. I can imagine this will cause a slight headache with pytest.

PR #742, opened 2021-11-05
gravel: check registry before deployment

jecluis:
With this patch set we now check whether a registry is reachable, and the container image exists at the registry, before allowing the node to deploy.

The major issue with this patch set is that it perpetuates the bad habit of not handling exceptions in an ideal way, conveying to the caller what might be wrong (aside from something being wrong). This, I believe, needs to be addressed as a whole in a future patch set.

### DIFF
--- a/src/gravel/controllers/containers.py
+++ b/src/gravel/controllers/containers.py
@@ -69,7 +69,7 @@ async def registry_check(registry: str, image: str, secure: bool) -> None:
     # check for an existing registry
     #
     try:
-        req = requests.get(f"{proto}://{registry}/v2/")
+        req = requests.get(f"{proto}://{registry}/v2/", timeout=5)
         if req.status_code != requests.codes.ok:
             raise UnknownRegistryError(
                 registry,
--- a/src/gravel/controllers/containers.py
+++ b/src/gravel/controllers/containers.py
@@ -8,8 +8,9 @@
 #
 import shlex
 from pathlib import Path
-from typing import Any, Dict, MutableMapping
+from typing import Any, Dict, MutableMapping, Optional
 
+import requests
 import toml
 
 from gravel.controllers.errors import GravelError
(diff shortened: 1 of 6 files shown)


src/gravel/controllers/containers.py
```diff
@@ -26,6 +27,60 @@ class ContainerPullError(ContainerError):
...
+        raise ImageNameError(msg=image)
+    (imgname, imgtag) = img
+
+    try:
+        req = requests.get(f"{proto}://{registry}/v2/{imgname}/tags/list")
+    except Exception as e:
+        raise ContainerError(msg=str(e))
+
+    if req.status_code != requests.codes.ok:
+        raise UnknownImageError(msg=imgname)
```
votdev, line 77: I prefer messages that are a little more meaningful. That's because if such an exception is processed by an API endpoint which simply converts it as a HTTPException with message `str(err)` this message is not really helpful to the user when it is shown in the UI.
```suggestion
        raise UnknownImageError(msg=f"The image '{imgname}' does not exist.")
```

src/gravel/controllers/containers.py
```diff
@@ -26,6 +27,60 @@ class ContainerPullError(ContainerError):
...
+    res = req.json()
+    assert res["name"] == imgname
+    if imgtag not in res["tags"]:
+        raise UnknownImageTagError(msg=image)
```
votdev, line 81: See above.
```suggestion
        raise UnknownImageTagError(msg=f"bla bla bla")
```

votdev (changes requested):

src/gravel/controllers/containers.py
jecluis: My only counter argument here is that the exception itself is already pretty self-explanatory, and the contents of the exception message should just clarify it. IMO, it's on the caller's plate to translate that to something that can be wired through to the API consumer.

Granted that's not happening at the moment, but it's beyond the scope of this PR.

jecluis:
@votdev pushed with suggested changes.

votdev (approved):
Tested and LGTM

PR #748, opened 2021-11-09
gravel: add network controller and API endpoints

tserong:
This adds the ability for gravel to read and write network interface configuration, nameservers and routes.  We've added an /api/local/network/config endpoint.  GET this, and you'll retrieve a dict of interfaces, a list of nameservers and a list of static routes.  In the simple case for a single NIC that's configured for DHCP, this'll give you:

```json
{
  "interfaces": {
     "eth0": {
      "name": "eth0",
      "config": {
        "bootproto": "dhcp",
        "ipaddr": "",
        "bonding_slaves": []
      }
    }
  },
  "nameservers": [],
  "routes": []
}
```

POST back to that same URL to apply a different config.  The config exposed is deliberately a subset of what's possible to specify in /etc/sysconfig/network/{config,routes,ifcfg-*,ifroute-*} files.  For 'interfaces', bootproto can be either "dhcp" or "static".  If it's "static", you also need to specify "ipaddr", and that needs to include the netmask or equivalent on the end (e.g.: "192.168.121.1/24"). If you want to create a bond, add a new interface to that dict and set bonding_slaves to a list of the physical interfaces to bond.  If the list is empty, it's just a normal interface.

'nameservers' is just a list of IP addresses, e.g.: ["8.8.8.8", ...]

'routes' is a list of dicts, each of which needs to include 'destination', 'gateway' and 'interface' as described in `man ifroute`.

When you apply the config, all network services are restarted.  Be particularly careful when applying a new config to the interface via which you're accessing the API.  Switch from DHCP to static IP and forget to configure DNS and/or a default route, and you can get yourself in trouble.  Likewise if you change the IP address to be on some other network.

Note that for NICs we've got an InterfaceModel which includes an InterfaceConfigModel.  This was based on Joao's work in https://github.com/aquarist-labs/aquarium/pull/603, the idea being that we may also want to add stats and other data to InterfaceModel. OTOH, that may duplicate information we already get from cephadm, so we might want to consider collapsing this instead.

## Checklist

---

<details>

- `jenkins run tumbleweed`
- `jenkins run leap`
- `jenkins run ubuntu`

</details>

### DIFF
--- a/src/gravel/controllers/resources/network.py
+++ b/src/gravel/controllers/resources/network.py
@@ -205,6 +205,7 @@ class Network(Ticker):
         return rawconf
 
     def _update_config(self, path: Path, new_config: Dict[str, str]):
+        new_config = new_config.copy()
         with path.open("r") as fin, atomic_write(path, overwrite=True) as fout:
             for line in fin:
                 stripped = line.strip()
@@ -214,7 +215,13 @@ class Network(Ticker):
                     if field in new_config:
                         # TODO: doesn't handle embedded quotes (but we're unlikely to care right now)
                         line = f'{field}="{new_config[field]}"\n'
+                        del new_config[field]
                 fout.write(line)
+            # This last bit adds any new fields that weren't already
+            # present to the end of the config file.
+            for field in new_config:
+                # (also doesn't handle embedded quotes, as above)
+                fout.write(f'{field}="{new_config[field]}"\n')
 
     def _update_route_file(
         self, path: Path, interface: str, routes: List[RouteModel]
--- a/src/gravel/controllers/resources/network.py
+++ b/src/gravel/controllers/resources/network.py
@@ -60,6 +60,12 @@ class RouteModel(BaseModel):
     interface: str = Field("-", title="Interface")
 
 
+class NetworkConfigModel(BaseModel):
+    interfaces: Dict[str, InterfaceModel]
+    nameservers: List[str]
+    routes: List[RouteModel]
+
+
 class Network(Ticker):
 
     _interfaces: Dict[str, InterfaceModel] = {}
@@ -223,6 +229,8 @@ class Network(Ticker):
             # No routes for this interface, delete config file if present
             path.unlink(missing_ok=True)
 
+    # TODO: This should probably just take a NetworkConfigModel,
+    # rather that separate interfaces, nameservers, routes.
     async def apply_config(
         self,
         interfaces: Dict[str, InterfaceModel],
--- a/src/gravel/controllers/resources/network.py
+++ b/src/gravel/controllers/resources/network.py
@@ -1,0 +1,297 @@
+# project aquarium's backend
+# Copyright (C) 2021 SUSE, LLC.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+import logging
+import os
+from pathlib import Path
+from typing import Dict, List, Optional
+
+from atomicwrites import atomic_write
+from fastapi.logger import logger as fastapi_logger
+from pydantic import BaseModel, Field
+
+from gravel.controllers.gstate import Ticker
+from gravel.controllers.utils import aqr_run_cmd
+
+logger: logging.Logger = fastapi_logger
+
+
+class InterfaceConfigModel(BaseModel):
+    # Simplified "UI friendly" version of the actual config, i.e. we're not
+    # exposing everything in /etc/sysconfig/network/ifcfg-*, only the bits that
+    # we want the user to be able to change.
+    # TODO: bootproto would be better as a StrEnum (so we could force "dhcp" or "static")
+    bootproto: str = Field("dhcp", title="Boot Protocol")
+    ipaddr: Optional[str] = Field("", title="IP Address")
+    # If bonding_slaves is _not_ empty, this interface is a bonding master.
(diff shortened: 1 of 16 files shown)


tserong:
I've now also added the ability to pass network config in to the /nodes/deployment/start and /nodes/join endpoints. I've tested this by manually hitting those endpoints with an appropriate request, e.g. for starting the deployment:

```
{
  "hostname": "node1",
  "network": {
    "interfaces": {
      "eth0": { 
        "name": "eth0",
        "config": {
          "bootproto": "static",
          "ipaddr": "192.168.121.89/24",
          "bonding_slaves": []
        }
      }
    },
    "nameservers": [
      "192.168.121.1"
    ],
    "routes": [
      {
        "destination": "default",
        "gateway": "192.168.121.1",
        "interface": "-"
      }
    ]
  },
  "ntpaddr": "pool.ntp.org",
  "registry": { 
    "registry": "registry.opensuse.org",
    "secure": true,
    "image": "filesystems/ceph/master/upstream/images/opensuse/ceph/ceph:latest"
  }
}
```

In the above case I'm applying a static IP address, but was careful to assign the same address as was already assigned by DHCP. If I'd chosen a different IP address, the browser would have lost its connection to the UI. This suggests that once network config is added to the setup wizard, we need some means of issuing a redirect if the IP address you're accessing aquarium on changes.

Anyway: the backend bits basically work, and there should, I hope, be enough in place, that the frontend bits can be done next if this PR passes review.

tserong:
(Happy to squash these commits, BTW)

src/gravel/controllers/resources/network.py
```diff
@@ -0,0 +1,305 @@
...
+        logger.debug(self._interfaces)
+        logger.debug(self._nameservers)
+        logger.debug(self._routes)
+
+    async def _should_tick(self) -> bool:
+        return not self._is_busy
+
+    def _refresh_interfaces(self) -> None:
+        # interfaces is the union of the physical interfaces (/sys/class/net/...) and
+        # any interfaces for which we have config files (/etc/sysconfig/network/ifcfg-*)
```
jhesketh, line 104: Is it correct to assume the config file in that position will be up to date? Would we be better off using `ip addr` and associated utils to get the information? Using the latter would also likely make it easier to move between wicked and networkmanager.

src/gravel/controllers/resources/network.py
```diff
@@ -0,0 +1,305 @@
...
+                config.bootproto = rawconf["BOOTPROTO"]
+
+            if iface not in interfaces:
+                interfaces[iface] = InterfaceModel(name=iface)
+            interfaces[iface].config = config
+
+        self._interfaces = interfaces
+
+    def _refresh_nameservers(self) -> None:
+        config = Path("/etc/sysconfig/network/config")
```
jhesketh, line 158: same question vs using other utilities (or /etc/resolv.conf etc).

src/gravel/controllers/resources/network.py
```diff
@@ -0,0 +1,305 @@
...
+            "NETCONFIG_DNS_STATIC_SERVERS" in rawconf
+            and rawconf["NETCONFIG_DNS_STATIC_SERVERS"] is not None
+        ):
+            self._nameservers = rawconf["NETCONFIG_DNS_STATIC_SERVERS"].split()
+        else:
+            self._nameservers = []
+
+    def _refresh_routes(self) -> None:
+        routes: List[RouteModel] = []
+        route_paths = [Path("/etc/sysconfig/network/routes")]
```
jhesketh, line 171: also from `ip route`?

src/gravel/controllers/resources/network.py
```diff
@@ -0,0 +1,305 @@
...
+    # rather that separate interfaces, nameservers, routes.
+    async def apply_config(
+        self,
+        interfaces: Dict[str, InterfaceModel],
+        nameservers: List[str],
+        routes: List[RouteModel],
+    ) -> None:
+        logger.debug("In Network.apply_config()")
+        self._is_busy = True
+        sysconfig = Path("/etc/sysconfig/network")
```
jhesketh, line 242: The `ip` utility can make these changes (afaik).

jhesketh:
Generally looks good to me, but wondering about the best way to read and apply the config (see inline). Sorry if it's already been discussed etc.

Your PR description would also make some helpful docs if you wanted to chuck that into git somewhere until we start writing real documentation.

jhesketh (dismissed):

tserong:
> Generally looks good to me, but wondering about the best way to read and apply the config (see inline). Sorry if it's already been discussed etc.

I forget whether or not this got discussed in detail anywhere, sorry :-) but the assumptions I was working from were:

- We'll always by using wicked, not networkmanager (because the former is the default on SUSE distros for servers, the latter for laptops/desktops)
- We only want to expose interfaces that physical exist (/sys/class/net/*/device is present) or are persistently configured (/etc/sysconfig/network/ifcfg-*)
- Nobody should be making changes to the network manually, outside of aquarium (nothing ephemeral that we'd need to pick up from `ip` invocations).

src/gravel/controllers/resources/network.py
tserong: With the default setup, /etc/resolv.conf is dynamically generated by netconfig based on variables in /etc/sysconfig/network/config, so it should be OK like this, provided my various assumptions are correct - see my other comment on that.

tserong:
Yikes ... resolving conflicts via the github web UI merges main into this branch. I don't like it - I'm gonna go rebase instead like I usually do.

jecluis (dismissed):
LGTM but we might want to update the config with new keys from the dict in `_update_config()`. I'll leave that to @tserong 's consideration. :)

tserong:
@jecluis I've got it adding new keys now in `_update_config()`. We may not strictly have needed that with the default setup, but it was an oversight on my part (thank you!), and realising that I figured it'd find a way to bite us later somehow if I didn't fix it ;-)

Mind a quick re-review? Also: want me to squash all these commits, or just merge as-is once approved?

jecluis (approved):

jecluis:
@tserong good to go if you are happy with it.

PR #752, opened 2021-11-10
gravel: clean up systemdisk

jecluis:
This is preparatory clean up work for a bigger cleanup and rework of how we deploy the system. In part to account for an inevitable change in how we install the system, but mostly (and more pressing) to better handle error states.

### DIFF
--- a/src/gravel/controllers/nodes/systemdisk.py
+++ b/src/gravel/controllers/nodes/systemdisk.py
@@ -136,6 +136,30 @@ class SystemDisk:
                 return True
         return False
 
+    async def exists(self) -> bool:
+        """Checks whether a System Disk exists in the system."""
+
+        lvmcmd = shlex.split("lvs --noheadings -o vg_name,lv_name @aquarium")
+        try:
+            out = await lvm(lvmcmd)
+        except LVMError:
+            return False
+
+        lst = [x.strip() for x in out.split("\n") if len(x) > 0]
+        if len(lst) == 0:
+            return False
+
+        lvs = [lv.split()[1] for lv in lst]
+        if "containers" not in lvs:
+            # systemdisk does not have a containers LV, we're in error state.
+            raise LVMError(msg="System Disk missing Containers LV.")
+        elif "systemdisk" not in lvs:
+            # systemdisk does not have a persistent state LV, we're in error
+            # state.
+            raise LVMError(msg="System Disk missing persistent state LV.")
+
+        return True
+
     async def create(self, gstate: GlobalState, devicestr: str) -> None:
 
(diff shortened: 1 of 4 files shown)


src/gravel/controllers/nodes/systemdisk.py
```diff
@@ -55,30 +55,64 @@ class OverlayError(GravelError):
...
+                except OverlayError as e:
+                    raise OverlayError(
+                        f"Unable to find overlay source for {dst}: {e.message}"
+                    )
+
+            lst.append(MountEntry(source=src, dest=dst, overlay=is_overlay))
     return lst
 
 
+async def lvm(args: str) -> str:
```
votdev, line 105: Isn't it better to use a `List[str]` here instead of using `shlex.split`? I think the developer better knows how to split the commands, especially when it comes to more complicated command args.

src/gravel/controllers/nodes/systemdisk.py
```diff
@@ -102,17 +136,34 @@ def mounted(self):
...
 
-    async def lvm(self, args: str) -> None:
-        cmd: List[str] = ["lvm"] + shlex.split(args)
-        retcode, _, err = await aqr_run_cmd(cmd)
+    async def exists(self) -> bool:
+        """Checks whether a System Disk exists in the system."""
+
+        lvmcmd = "lvs --noheadings -o vg_name,lv_name @aquarium"
+        try:
+            out = await lvm(lvmcmd)
```
votdev, line 144: I think it should be something like that:
```suggestion
            out = await lvm(["lvs", "--noheadings", "-o", "vg_name,lv_name", "@aquarium"]])
```

votdev (dismissed):

src/gravel/controllers/nodes/systemdisk.py
mgfritch: I'd also prefer using an argv rather than shelx ..

mgfritch (approved):
lgtm, minus the change requested from @votdev

src/gravel/controllers/nodes/systemdisk.py
jecluis: While I understand the point of getting a list of strings, why would manually passing the list of strings be better than let the shell lexicon parser, or more reliable?

jhesketh (commented):
Seems reasonable to me (I have no opinion on shelx vs argv).

tserong (approved):
Approving so #757 can be rebased on top of this

jecluis:
The function now takes a list, but the code is still relying on `shlex` because I think that's less error prone.

PR #756, opened 2021-11-29
glass: add row selection to datatable

s0nea:

## Checklist

---

<details>

- `jenkins run tumbleweed`
- `jenkins run leap`
- `jenkins run ubuntu`

</details>

### DIFF
--- a/src/glass/src/app/shared/components/datatable/datatable.component.ts
+++ b/src/glass/src/app/shared/components/datatable/datatable.component.ts
@@ -28,7 +28,10 @@ import * as _ from 'lodash';
 import { Subscription, timer } from 'rxjs';
 
 import { Icon } from '~/app/shared/enum/icon.enum';
-import { DatatableColumn } from '~/app/shared/models/datatable-column.type';
+import {
+  DatatableCellTemplateName,
+  DatatableColumn
+} from '~/app/shared/models/datatable-column.type';
 import { DatatableData } from '~/app/shared/models/datatable-data.type';
 
 export enum SortDirection {
@@ -48,6 +51,8 @@ export class DatatableComponent implements OnInit, OnDestroy {
   checkIconTpl?: TemplateRef<any>;
   @ViewChild('yesNoIconTpl', { static: true })
   yesNoIconTpl?: TemplateRef<any>;
+  @ViewChild('rowSelectTpl', { static: true })
+  rowSelectTpl?: TemplateRef<any>;
   @ViewChild('actionMenuTpl', { static: true })
   actionMenuTpl?: TemplateRef<any>;
   @ViewChild('mapTpl', { static: true })
@@ -85,8 +90,25 @@ export class DatatableComponent implements OnInit, OnDestroy {
   @Input()
   autoReload: number | boolean = 15000;
 
+  // Row property used as unique identifier for the shown data. Only used if
+  // the row selection is enabled. Will throw an error if property not found
+  // in given columns. Defaults to 'id'.
+  @Input()
+  identifier = 'id';
+
+  // Defines the following row selection types:
+  // none: no row selection
+  // single: allows single-select
+  // multi: allows multi-select
+  // Defaults to no row selection.
+  @Input()
+  selectionType: 'single' | 'multi' | 'none' = 'none';
+
   @Output()
   loadData = new EventEmitter();
+
+  @Output()
+  selectionChange = new EventEmitter<DatatableData[]>();
 
   // Internal
   public icons = Icon;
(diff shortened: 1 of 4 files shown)


src/glass/src/app/shared/components/datatable/datatable.component.ts
```diff
@@ -221,6 +257,28 @@ export class DatatableComponent implements OnInit, OnDestroy {
     this.reloadData();
   }
 
+  onSelectionChange($event: any, row: DatatableData) {
+    if ($event.target.checked) {
+      this.selected.push(row);
+    } else {
+      const selectedIndex = _.map(this.selected, this.identifier).indexOf(row[this.identifier]);
```
votdev, line 264: You can also use the following code which looks more intuitive IMO.
```suggestion
      const selectedIndex = _.findIndex(this.selected, [this.identifier, row[this.identifier]]);
```

votdev (changes requested):

votdev:
Working on the new installer i realized that we need a flag to tell the datatable which selection mode to use: none, single or multiple.

s0nea:
jenkins run leap

src/glass/src/app/shared/components/datatable/datatable.component.ts
```diff
@@ -85,6 +91,16 @@ export class DatatableComponent implements OnInit, OnDestroy {
...
 
+  // Row property used as unique identifier for the shown data. Only used if
+  // the row selection is enabled. Will throw an error if property not found
+  // in given columns. Defaults to 'id'.
+  @Input()
+  identifier = 'id';
+
+  // Defines the row selection type. Defaults to no row selection.
+  @Input()
+  selectionType: DatatableSelectionType = DatatableSelectionType.none;
```
votdev, line 102: Using an enum as input property is not so good IMO because in the HTML code you can not use it. With the current implementation it has to look like:
```
      <glass-datatable [data]="devices"
                       sortHeader="path"
                       selectionType="1"
                       [columns]="devicesColumns">
      </glass-datatable>
```
Better would be
```
      <glass-datatable [data]="devices"
                       sortHeader="path"
                       selectionType="single"
                       [columns]="devicesColumns">
      </glass-datatable>
```

```suggestion
  selectionType: 'none' | 'single' | 'multi' = 'none';
```

votdev (changes requested):

src/glass/src/app/shared/components/datatable/datatable.component.ts
s0nea: Yes, if you're going to use it in the HTML file, you're right. But there is also the possibility to define variable in the *.ts file, e.g.:

```selectionType: DatatableSelectionType = DatatableSelectionType.single; ```

and then add it to the HTML file:

```[selectionType]="selectionType"```

But well... for me both options are fine.

**Comment by votdev** [comment]
I realized that a `selectionChange` event is missing. It's a simple implementation, but for the moment this is ok.
```
diff --git a/src/glass/src/app/shared/components/datatable/datatable.component.ts b/src/glass/src/app/shared/components/datatable/datatable.component.ts
--- a/src/glass/src/app/shared/components/datatable/datatable.component.ts	(revision 94281fbc3f42452f200192ff7706816ea333ba04)
+++ b/src/glass/src/app/shared/components/datatable/datatable.component.ts	(date 1638448078860)
@@ -104,6 +104,9 @@
   @Output()
   loadData = new EventEmitter();
 
+  @Output()
+  selectionChange = new EventEmitter<DatatableData[]>();
+
   // Internal
   public icons = Icon;
   public page = 1;
@@ -285,6 +288,7 @@
         this.selected.splice(selectedIndex, 1);
       }
     }
+    this.selectionChange.emit(this.selected);
   }
 
   private updateSelection(): void {
```

src/glass/src/app/shared/components/datatable/datatable.component.ts
votdev: > Yes, if you're going to use it in the HTML file, you're right. But there is also the possibility to define variable in the *.ts file

Right, but IMO this is unnecessary code because this type of configuration is fixed in nearly 99.9%, thus setting the attribute in the HTML code is the way to go.

> Btw: we do have the same for sortDirection:

Hmmm, right. IMO we should also cleanup this issue. It's IMO not the way a developer is using this component and attribute.

votdev:
I think we also need a checkbox in the column header to select all rows in one step. Imagine the user needs to select the disk devices in the installer step that are used for storage. In a production environment this could be 50, 100 or more devices. IMO we should implement that in a separate PR.

votdev (approved):

s0nea:
> I think we also need a checkbox in the column header to select all rows in one step. Imagine the user needs to select the disk devices in the installer step that are used for storage. In a production environment this could be 50, 100 or more devices. IMO we should implement that in a separate PR.

Yep, I already thought about the same. I'm in favor of a separate PR as well.

PR #757, opened 2021-11-29
gravel: deployment rework

jecluis:
This patch set reworks how we do the install and deployment phase for seed nodes and nodes joining the network.

While it works as it is, all unit tests are broken. This needs considerable validation still. And there are known, minor bugs.

Additionally, requires a rebase on top of current `main`.

This work won't be complete until the frontend is also adjusted.

Consider this a draft, up for review up until this point.


src/gravel/api/local.py
```diff
@@ -132,7 +135,9 @@ async def get_inventory(
     response_model=NodeStatusReplyModel,
 )
 async def get_status(
-    request: Request, _: Callable = Depends(jwt_auth_scheme)
+    request: Request,
+    jwt: Any = Depends(jwt_auth_scheme),
+    gate: Any = Depends(install_gate),
```
votdev, line 140: Remove this dependency. The API endpoint is called by the installed on the welcome page where the node is not installed right now.

src/aquarium.py
```diff
@@ -81,19 +89,99 @@ async def aquarium_startup(_: FastAPI, aquarium_api: FastAPI):
...
     await gstate.start()
 
-    # Add instances into FastAPI's state:
-    aquarium_api.state.gstate = gstate
-    aquarium_api.state.nodemgr = nodemgr
+    logger.info("Post-Init Deployment.")
+    deployment.postinit(gstate, nodemgr)
+
+    app.state.gstate = gstate
+    app.state.nodemgr = nodemgr
```
votdev, line 140: This line needs to be moved to line 110 right after `app.state.deployment = deployment` and BEFORE the `while not _shutting_down and not deployment.installed:` loop.

src/gravel/controllers/deployment/mgr.py
```diff
@@ -0,0 +1,635 @@
...
+            logger.info("Node is currently installing; no op.")
+            return
+
+        assert self._task_main is not None
+        assert not self._task_main.done()
+        assert self._task_install is None
+
+        # reset error, if any
+        self._error = DeploymentErrorModel()
+        self._task_install = asyncio.create_task(self._install_task(device))
```
votdev, line 486: Exceptions thrown in the task are not catched, thus executing `/api/deploy/install` with an device that does not exist will always return `{success:true}`.

votdev:
The deployment fails with the following error message:
```
INFO:     2021-12-03 10:59:55 -- create -- Bootstrap complete with success.
DEBUG:    2021-12-03 10:59:58 -- create -- Create progress: 85%, Configuring deployment.
ERROR:    2021-12-03 11:01:10 -- ceph -- error running command: rc = -2, reason = all mgr daemons do not support module 'bubbles', pass --force to force enablement
ERROR:    2021-12-03 11:01:10 -- ceph -- unable to enable module bubbles: all mgr daemons do not support module 'bubbles', pass --force to force enablement
```

jecluis:
> The deployment fails with the following error message:
> 
> ```
> INFO:     2021-12-03 10:59:55 -- create -- Bootstrap complete with success.
> DEBUG:    2021-12-03 10:59:58 -- create -- Create progress: 85%, Configuring deployment.
> ERROR:    2021-12-03 11:01:10 -- ceph -- error running command: rc = -2, reason = all mgr daemons do not support module 'bubbles', pass --force to force enablement
> ERROR:    2021-12-03 11:01:10 -- ceph -- unable to enable module bubbles: all mgr daemons do not support module 'bubbles', pass --force to force enablement
> ERROR:    2021-12-03 11:04:10 -- create -- Unable to start Bubbles.
> ERROR:    2021-12-03 11:04:16 -- create -- Create error: Failed configuring the deployment.
> ```

This is likely because you're using the default registry and image, which currently doesn't support bubbles. For this to actually succeed you need to provide a custom registry with a bubbles container.

src/gravel/controllers/deployment/mgr.py
jecluis: That's correct. We're not throwing any exceptions through that function, just setting error messages in case there's an error. The only way to properly deal with that is by calling the `/api/deploy/install` endpoint and then check the `/api/deploy/status` endpoint for errors.

tserong (commented):
-ETOOBIGTOREVIEW ;-)

I have actually read all the commits (which are at least each nice and reasonably self contained / small enough to consume one at a time), but at a certain point I still hit that "eyes glaze over" problem. Also I haven't attempted any testing.

My main takeaway here is that as the system disk setup is now split out, the UI first has to hit "/api/deploy/install" passing in the system disk to use, then later hits either "/api/deploy/create" or "/api/deploy/join" (whereas before everything was done when the UI hit one of those last two endpoints).

How will this intersect with imaging the system disk from the live image thingy @AvengerMoJo is working on? I guess that'll somehow slot in somewhere near `SystemDisk.create()`?

Also I noticed a tiny discrepancy in some field names, which is probably not actually terribly important. Funny what you pick up on sometimes...

jecluis:
> How will this intersect with imaging the system disk from the live image thingy @AvengerMoJo is working on? I guess that'll somehow slot in somewhere near `SystemDisk.create()`?

Yeah, that's the plan.

votdev (approved):

mgfritch (commented):
any chance we can fix the unit test / lint failures? otherwise, seems reasonable to get this in at this point!

mgfritch (approved):

PR #765, opened 2022-01-07
doc: Support Assessment Dashboard Infrastructure Design

irq0:
Signed-off-by: Marcel Lauhoff <marcel.lauhoff@suse.com>


doc/spec/support-assessment.md
```diff
@@ -0,0 +1,237 @@
...
+Proposed. Prototype demoed.
+
+## Milestones
+
+1. PoC and Design Doc
+2. PoC -> MVP: Code in repository, deployable, reasonable test coverage, initial set of rules
+   written, basic dashboard contain {red, amber, green} and 'why?'
+1337. Profit
+
+## Discussions and Ideas
```
irq0, line 183: Better put this elsewhere (which would be where?)?

AvengerMoJo: Create a confluence page?

doc/spec/support-assessment.md
```diff
@@ -0,0 +1,237 @@
...
+Where we create new exporters we may want to consider adding
+additional `/detail/$metric_name`-style endpoints. Bubbles could query
+them to receive additional information for human consumption not
+available from Prometheus directly.
+
+## Alternatives
+
+### Cephadm Configchecks
+
+Configuration checks build into the cephadm manager module. Tightly
```
asettle, line 120: In ceph/ceph or aquarium?

doc/spec/support-assessment.md
```diff
@@ -0,0 +1,237 @@
...
+too risky.
+
+Data gathering is also challenging: Sources may change formats,
+joining is not always straight forward because of different
+naming schemes. Keeping the memory footprint low and supporting
+historical queries is challenging.
+
+## Why Prometheus?
+
+Prometheus a de facto industry standard time series database and
```
asettle, line 154:
```suggestion
Prometheus is a de facto industry standard time-series database and

File: doc/spec/support-assessment.md [hunk 4]
```diff
@@ -0,0 +1,237 @@
...
+Going with Prometheus saves us from building our own rule engine or
+writing a lot of code to process state into alert conditions
+(configchecks.py). Writing rules will become easier the more we have
+and the more data sources we integrate. We will loose some flexibility, as
+we can't "just" query arbitrary string state data without processing
+it first.
+
+## Status
+
+Proposed. Prototype demoed.
```
asettle, line 174: Let's add a link here and make it public. @jecluis did you save the recordings or was that @ml8mr ?

doc/spec/support-assessment.md
asettle: Here for sure: https://github.com/aquarist-labs/forum

asettle: To be clearer: https://github.com/aquarist-labs/forum/discussions

doc/spec/support-assessment.md
```diff
@@ -0,0 +1,237 @@
...
+
+- When gathering and joining data we must take into account that
+  keeping a copy of all state in memory won't scale
+- We must join data from diverse sources like the manger, Prometheus,
+  cluster maps, Aquarium, Ceph orchestrator. All having different schema.
+- Rules are subject to configuration and change. Changes here must be easy.
+
+## Proposed Change
+
+Use Prometheus to gather, prepare, join and evaluate rules on numeric
```
l-mb, line 28: My question is - if we convert everything to a metric Prometheus can interact with, does this mean we lose fidelity for reporting issues to users later, since we may no longer have the original datum to report?

doc/spec/support-assessment.md
```diff
@@ -0,0 +1,237 @@
...
+metrics. Rules are Prometheus alerts, written in PromQL and tested
+with the [`promtool`](https://prometheus.io/docs/prometheus/latest/configuration/unit_testing_rules)
+
+Begin with readily-available metrics from Prometheus node exporter and
+Ceph manager's Prometheus exporter. Implement custom Prometheus
+exporters where necessary. Abstract complex queries behind simple
+numeric metrics where appropriate and necessary due to the lack of a
+string metric type. Example: Metric `outdated_client 42`
+instead of metrics `client{version="23.42"} 42`.
```
l-mb, line 37: This is one such example, since it'd be useful to be able to report which clients run which versions; it could be fixed by tracking this as a BCD number (so we can have our own 2K22 bug at some point), or some other scheme that doesn't lose information.

doc/spec/support-assessment.md
irq0: https://github.com/aquarist-labs/forum/discussions/24

doc/spec/support-assessment.md
```diff
@@ -0,0 +1,237 @@
...
+
+Rules on data from logs and audit events will require another approach
+and are out of scope of this change.
+
+A Bubbles extra manages and configures Prometheus and collects alerts
```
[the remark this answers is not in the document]
irq0: Bubbles has an extension mechanism. Modules inside modules == extra :rofl:. 
https://github.com/ceph/bubbles/tree/main/extras

I'll try to improve clarity

doc/spec/support-assessment.md
irq0: ceph/ceph. This thing here: https://github.com/ceph/ceph/blob/master/src/pybind/mgr/cephadm/configchecks.py

doc/spec/support-assessment.md
irq0: Sometimes we do, yes. I expect this mostly in cases where we have to hide complex or string metric processing behind gauges/counters. To get more detail when something fires we could add another endpoint to the exporters to be fetched when a user hits the dashboard. 

To make it more concrete an Alertmanager rule could look like this:
```
  - alert: OutdatedClients
    expr: outdated_clients > 0
    for: 23m
    labels:
      severity: warning
    annotations:
      summary: Cluster has {{ $value }} outdated clients
      details: https://@@VERSION_EXPORTER@@/details
```

(Where @@VERSION_EXPORTER@@ is something we resolve from our cluster state)

I've seen people add *dashboard* annotations pointing to things like Grafana. This kind of goes in that direction.

doc/spec/support-assessment.md
irq0: Do I grok this correctly, encode the version number as a gauge metric?

Something like this?
`client{type="cephfs", name="somehorst.example.com",pid="2342",implementation="libcephfs"} BCD_ENCODED_VERSION_NUMBER`

PR #770, opened 2022-01-12
glass/gravel: Add button to redirect to Bubbles UI

votdev:
Peek 2022-01-12 18-48

Make it as easy as possible for the user to switch to the Bubbles UI. Note, the Bubbles UI will not have a button to switch to Aquarium because Bubbles does not know anything about Aquarium.

## Checklist

---

<details>

- `jenkins run tumbleweed`
- `jenkins run leap`
- `jenkins run ubuntu`

</details>

### DIFF
--- a/src/glass/src/app/shared/services/api/status.service.ts
+++ b/src/glass/src/app/shared/services/api/status.service.ts
@@ -12,28 +12,61 @@
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
  */
+/* eslint-disable @typescript-eslint/naming-convention */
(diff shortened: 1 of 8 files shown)


src/glass/src/app/shared/services/api/status.service.ts
```diff
@@ -12,28 +12,61 @@
...
   checks: { [id: string]: HealthCheckSummary };
-}
+};
+
+export type PGState = {
+  state_name: string;
+  count: number;
+};
+
+export type PGMap = {
```
s0nea, line 35: Just out of curiosity: how do these values relate to the PR?

jecluis (approved):

